PR #9, opened 2022-12-01
local development on windows

olevett:
- add a gitattributes file to make sure line endings are reasonable
- add a dockerignore file to avoid sending some large videos into the build context


Mjwillis (approved):
LGTM

PR #11, opened 2022-12-06
feat: add initial test GitHub action

olevett:


olevett:
The remote FTP test hangs on GitHub actions so I've disabled it to start with, but there's probably a better option about it, if we think the test is necessary/worth including

Mjwillis (approved):
LGTM

PR #23, opened 2022-12-08
feat: Add github actions to build and push to ECR image

oskarwilliams378:
Add github actions to build and push to ECR image.


.github/workflows/application.yml
```diff
@@ -0,0 +1,47 @@
...
+  build-and-push-image:
+    if: github.event_name == 'push' && github.ref_name == 'main'
+    strategy:
+      matrix:
+        env: [dev, poc]
+    name: Build and push docker image for "${{ matrix.env }}" environment to ECR
+    needs: test
+    uses: ./.github/workflows/build-and-push.yml
+    with:
+      push: true
```
olevett, line 42: Is it simpler to calculate the boolean here from the if block above, and then just have a "build-and-push" that may or may not actually push, at this level?

olevett (approved):

.github/workflows/application.yml
oskarwilliams378: Done

PR #29, opened 2022-12-09
feat: Update AppRunner configuration

olevett:
- isn't abailable in eu-west-2, so this adds an eu-west-1 provider
- ensure the lev service is proxied correctly in ECR


olevett:
Because of the dependency on the ECR configuration, which is managed through the demo environment, this may need to be deployed to demo before dev.

oskarwilliams378 (approved):

PR #30, opened 2022-12-13
chore: Remove redundant plan checking

oskarwilliams378:
Remove redundant plan checking as we do not halt after plan anyways


Mjwillis (approved):
LGTM

Issue #22, opened 2022-12-08
Deploy POC to ECS


olevett:
As part of the infrastructure build out, we need to deploy the POC Kotlin/Docker project to ECS, using an ECR registry

PR #37, opened 2022-12-13
22 terraform ecs

oskarwilliams378:
#22 
Add terraform to deploy:
ECS
LB
CloudFront
WAF

and subsequent dependencies


terraform/demo/main.tf
```diff
@@ -48,3 +62,16 @@ module "lev_api" {
   }
   environment_name = "demo"
 }
+
+
+module "data-share-service" {
+  source = "../modules/data-share-service"
+  providers = {
+    aws.us-east-1 = aws.us-east-1
```
olevett, line 70: We presumably need this for CloudFront?

terraform/demo/main.tf
```diff
@@ -48,3 +62,16 @@ module "lev_api" {
...
+  }
+  environment                 = "demo"
+  ecr_url                     = "${data.aws_caller_identity.current.account_id}.dkr.ecr.eu-west-2.amazonaws.com"
+  service_port                = 8080
+  cloudwatch_retention_period = 30
+  vpc_cidr                    = "10.158.16.0/20"
```
olevett, line 76: is this just a random choice?

terraform/demo/main.tf
```diff
@@ -48,3 +62,16 @@ module "lev_api" {
...
 }
+
+
+module "data-share-service" {
+  source = "../modules/data-share-service"
+  providers = {
+    aws.us-east-1 = aws.us-east-1
+  }
+  environment                 = "demo"
+  ecr_url                     = "${data.aws_caller_identity.current.account_id}.dkr.ecr.eu-west-2.amazonaws.com"
```
olevett, line 73: I _think_ I'd like this to come out of the ecr module as an output, but lets not worry about it for now (it'll make the non-demo environment trickier anyway)

terraform/demo/.terraform.lock.hcl
```diff
@@ -1,6 +1,24 @@
...
+    "zh:841fc4ac6dc3479981330974d44ad2341deada8a5ff9e3b1b4510702dfbdbed9",
+    "zh:91be62c9b41edb137f7f835491183628d484e9d6efa82fcb75cfa538c92791c5",
+    "zh:acd5f442bd88d67eb948b18dc2ed421c6c3faee62d3a12200e442bfff0aa7d8b",
+    "zh:ad5720da5524641ad718a565694821be5f61f68f1c3c5d2cfa24426b8e774bef",
+    "zh:e63f12ea938520b3f83634fc29da28d92eed5cfbc5cc8ca08281a6a9c36cca65",
+    "zh:f6542918faa115df46474a36aabb4c3899650bea036b5f8a5e296be6f8f25767",
+  ]
+}
+
 provider "registry.terraform.io/hashicorp/aws" {
```
olevett, line 22: Is there a reason we don't have the random provider in a lock here?

terraform/modules/data-share-service/cloudfront.tf
```diff
@@ -0,0 +1,66 @@
...
+      # AWS enforce a maximum of 60s, but we can request more if desired.
+      # See https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/distribution-web-values-specify.html#DownloadDistValuesOriginResponseTimeout
+      origin_read_timeout = 60
+    }
+  }
+
+  enabled         = true
+  is_ipv6_enabled = false
+
+  #  logging_config {
```
olevett, line 23: Do we have a ticket/something tracking this logging set up?

terraform/modules/data-share-service/ecs-security.tf
```diff
@@ -0,0 +1,23 @@
...
+  vpc_id      = module.vpc.vpc_id
+
+  ingress {
+    protocol        = "tcp"
+    from_port       = var.service_port
+    to_port         = var.service_port
+    security_groups = aws_security_group.lb_auto.*.id
+  }
+
+  egress {
```
olevett, line 13: This is a broad "allow" right? I think that's fine for POC, but we should lock this down as and when things have stabilised and we know what legit outbound traffic is.

terraform/modules/data-share-service/ecs.tf
```diff
@@ -0,0 +1,60 @@
+resource "aws_ecs_cluster" "main" {
+  name = var.environment
+}
+
+resource "aws_ecs_task_definition" "gdx_data_share_poc" {
+  family                   = "${var.environment}-gdx-data-share-poc"
+  requires_compatibilities = ["FARGATE"]
+  network_mode             = "awsvpc"
+  memory                   = 512
+  cpu                      = 256
```
olevett, line 10: This seems fairly low for a JVM service, but I guess we'll have to wait and see

terraform/demo/main.tf
oskarwilliams378: Yep

terraform/demo/main.tf
oskarwilliams378: Yeah tracking that across to non demo is an issue.

Also all it is is the account number followed by a constant string, so it is less important to output

terraform/demo/.terraform.lock.hcl
oskarwilliams378: I haven't init'ed there, done

terraform/demo/main.tf
olevett: I think we should be able to do some remote state/similar, but lets not worry for now

terraform/modules/data-share-service/cloudfront.tf
oskarwilliams378: Added logging now

terraform/modules/data-share-service/ecs-security.tf
oskarwilliams378: Other services I have seen have been perfectly happy with this, and it is the default with AWS, but am happy to lock down later

terraform/modules/data-share-service/ecs.tf
oskarwilliams378: Yeah I can imagine so, I just left this as a start point

terraform/demo/main.tf
```diff
@@ -48,3 +62,16 @@ module "lev_api" {
   }
   environment_name = "demo"
```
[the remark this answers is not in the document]
oskarwilliams378: Done

olevett (approved):

Issue #34, opened 2022-12-13
SQS Queues for service


olevett:
The readme defines a variety of topics and queues required for the service  - https://github.com/alphagov/gdx-data-share-poc#environment-variables---queues-and-topics

We need to 
- implement these as terraform config
- pass these into the ECS service

PR #45, opened 2022-12-14
feat:  Add initial SQS queues

olevett:
further work on #34

### DIFF
--- a/terraform/modules/data-share-service/sqs.tf
+++ b/terraform/modules/data-share-service/sqs.tf
@@ -1,23 +1,23 @@
 module "data_processor_queue" {
   source      = "../sqs"
-  queue_name        = "${var.environment}-gdx-data-share-poc-data-processor"
+  queue_name  = "${var.environment}-gdx-data-share-poc-data-processor"
   environment = var.environment
 }
 
 module "audit_queue" {
   source      = "../sqs"
-  queue_name        = "${var.environment}-gdx-data-share-poc-audit-queue"
+  queue_name  = "${var.environment}-gdx-data-share-poc-audit-queue"
   environment = var.environment
 }
 
 module "outbound_adaptor_queue" {
   source      = "../sqs"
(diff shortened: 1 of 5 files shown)


olevett:
We don't directly actually need the DLQ for the Audit flow, and as per #46 , I think there's a neater long term way of doing this.

terraform/modules/data-share-service/sqs.tf
```diff
@@ -0,0 +1,23 @@
+module "data_processor_queue" {
+  source      = "../sqs"
+  queue_name  = "${var.environment}-gdx-data-share-poc-data-processor"
```
oskarwilliams378, line 3: the others all end in queue?

terraform/modules/sqs/main.tf
```diff
@@ -0,0 +1,69 @@
+data "aws_caller_identity" "current" {}
+
+# rework as part of #46
+#tfsec:ignore:aws-iam-no-user-attached-policies 
+resource "aws_iam_user" "sqs_user" {
+  name = "${var.environment}-sqs-queue-${var.queue_name}"
```
oskarwilliams378, line 6: is this going to have the env in twice then? maybe just do 
```suggestion
  name = "${var.queue_name}-sqs-queue-user"
```

terraform/modules/sqs/main.tf
```diff
@@ -0,0 +1,69 @@
...
+  }
+}
+
+resource "aws_iam_user_policy" "sqs_policy" {
+  name   = "{var.environment}-sqs-queue-${var.queue_name}"
+  policy = data.aws_iam_policy_document.sqs_policy.json
+  user   = aws_iam_user.sqs_user.name
+}
+
+resource "aws_kms_key" "kms_key" {
```
oskarwilliams378, line 34: aliases maybe? (see #48 )

oskarwilliams378 (approved):
couple v minor things, not blocking

PR #48, opened 2022-12-14
chore: Add aliases for kms keys

oskarwilliams378:


terraform/modules/data-share-service/kms.tf
```diff
@@ -1,10 +1,15 @@
 resource "aws_kms_key" "log_key" {
-  description         = "Logs encryption key"
+  description         = "Key used to encrypt cloudfront and cloudwatch logs"
   enable_key_rotation = true
 
   policy = data.aws_iam_policy_document.kms_log_key_policy.json
 }
 
+resource "aws_kms_alias" "log_key_alias" {
+  name          = "alias/${var.environment}-log-key"
```
olevett, line 9: Do we want a structure like `/alias/env/purpose` rather than alias/env-variable

olevett:
Minor comment, but happy to proceed

olevett (approved):

terraform/modules/data-share-service/kms.tf
oskarwilliams378: oh yeah, better if possible

PR #52, opened 2022-12-14
feat: Setup codedeploy deploy

oskarwilliams378:


.github/workflows/trigger_codedeploy.yml
```diff
@@ -0,0 +1,104 @@
+name: Deploy
+
+on:
+  workflow_call:
+    inputs:
+      environment:
+        required: true
+        type: string
+      ecs_task_definition_filepath:
```
olevett, line 9: Are these things just internal to this step? If so, do we need to pass them in or can we hard code them>

olevett (approved):

PR #55, opened 2022-12-14
feat: Add first pass at RDS configuration for POC

### LINKED ISSUE #33: Aurora postgres for configuration
We need to set up a Postgres compatible Aurora database to be used as a backing store for some of the system configuration. This is currently configured only in the [application-dev.yml](https://github.com/alphagov/gdx-data-share-poc/blob/58317b0c1063c998c5d226bf530888eddd7f6e6c/src/main/resources/application-dev.yml) config.

Ideally, we don't want to be passing around raw credentials, either using AWS managed parameters/secrets, or using IAM.

Note we need both `flyway` and `r2dbc` config as flyway manages migrations and r2dbc is used for application code.

olevett:
Resolves #33, but as with SNS/SQS, I think we'll probably want to dive into using IAM/equivalents over this direct password passing

### DIFF
--- a/terraform/modules/data-share-service/ecs.tf
+++ b/terraform/modules/data-share-service/ecs.tf
@@ -53,7 +53,14 @@ resource "aws_ecs_task_definition" "gdx_data_share_poc" {
         { "name" : "HMPPS_SQS_QUEUES_ODG_QUEUE_NAME", "value" : module.other_department_queue.queue_name },
         { "name" : "HMPPS_SQS_QUEUES_ODG_DLQ_ACCESS_KEY_ID", "value" : module.other_department_queue.dead_letter_queue_access_key_id },
         { "name" : "HMPPS_SQS_QUEUES_ODG_DLQ_SECRET_ACCESS_KEY", "value" : module.other_department_queue.dead_letter_queue_access_key_secret },
-        { "name" : "HMPPS_SQS_QUEUES_ODG_DLQ_NAME", "value" : module.other_department_queue.dead_letter_queue_name }
+        { "name" : "HMPPS_SQS_QUEUES_ODG_DLQ_NAME", "value" : module.other_department_queue.dead_letter_queue_name },
+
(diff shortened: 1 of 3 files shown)


terraform/modules/data-share-service/ecs.tf
```diff
@@ -53,7 +53,14 @@ resource "aws_ecs_task_definition" "gdx_data_share_poc" {
         { "name" : "HMPPS_SQS_QUEUES_ODG_QUEUE_NAME", "value" : module.other_department_queue.queue_name },
         { "name" : "HMPPS_SQS_QUEUES_ODG_DLQ_ACCESS_KEY_ID", "value" : module.other_department_queue.dead_letter_queue_access_key_id },
         { "name" : "HMPPS_SQS_QUEUES_ODG_DLQ_SECRET_ACCESS_KEY", "value" : module.other_department_queue.dead_letter_queue_access_key_secret },
-        { "name" : "HMPPS_SQS_QUEUES_ODG_DLQ_NAME", "value" : module.other_department_queue.dead_letter_queue_name }
+        { "name" : "HMPPS_SQS_QUEUES_ODG_DLQ_NAME", "value" : module.other_department_queue.dead_letter_queue_name },
+
+        { "name" : "SPRING_FLYWAY_URL", "value" : "jdbc:postgresql://${aws_rds_cluster.rds_postgres_cluster.endpoint}" },
```
olevett, line 58: This format may well be wrong, but I want to get _something_ up to iterate on

terraform/modules/data-share-service/rds.tf
```diff
@@ -0,0 +1,35 @@
...
+resource "random_password" "rds_password" {
+  length  = 16
+  special = false
+}
+
+resource "aws_kms_key" "rds_key" {
+  description         = "Key for RDS encryption"
+  enable_key_rotation = true
+}
+
```
oskarwilliams378, line 15: I was having a think about this, should we keep all the KMS keys in the KMS file?

oskarwilliams378 (approved):
One minor comment I am not sure about :)

PR #93, opened 2022-12-19
feat: Update oauth flow for management client

oskarwilliams378:
Update oauth flow for management client


oskarwilliams378:
THis should stay as implicit

PR #96, opened 2022-12-20
Db rework

Mjwillis:
https://trello.com/c/GvMczKDl

This ticket is to make the DB structure more flexible for publishers and consumers.  Each have subscriptions to pub or consume.  

The types of events are explicit in the db
The dataset (or structures) are also defined

Removed the FTP part and just as a hook where we could send to S3 or POST to an endpoint.


src/main/kotlin/uk/gov/gdx/datashare/repository/PublisherSubscriptionRepository.kt
```diff
@@ -0,0 +1,12 @@
+package uk.gov.gdx.datashare.repository
+
+import org.springframework.data.r2dbc.repository.Query
+import org.springframework.data.repository.kotlin.CoroutineCrudRepository
+import org.springframework.stereotype.Repository
+
+@Repository
+interface PublisherSubscriptionRepository : CoroutineCrudRepository<PublisherSubscription, Long> {
```
olevett, line 8: PublisherSubscription feels slightly awkward naming wise, but I don't have much better - I guess it's probably more `EventSubscription`, because events could conceivably come from multiple publishers?

src/main/kotlin/uk/gov/gdx/datashare/service/DataReceiverService.kt
```diff
@@ -33,25 +34,37 @@ class DataReceiverService(
...
+    val subscription = publisherSubscriptionRepository.findByClientIdAndEventType(
+      authenticationFacade.getUsername(),
+      eventPayload.eventType.toString()
+    ) ?: throw RuntimeException("Publisher ${authenticationFacade.getUsername()} does not have a subscription for event type ${eventPayload.eventType}")
 
-    // check if client is allowed to send this type of event
-    if (dataProvider.eventType != eventPayload.eventType.toString()) {
-      throw RuntimeException("Client ${dataProvider.clientName} is not allowed to provide ${eventPayload.eventType} events")
+    val dataSet = eventDatasetRepository.findById(subscription.datasetId)
+      ?: throw RuntimeException("Client ${authenticationFacade.getUsername()} is not a known dataset")
```
olevett, line 43: This message doesn't quite tie together, I'd also be curious if there's any risk of leaking more data than we should through here

src/main/resources/db/migration/V1_4__new_schema.sql
```diff
@@ -0,0 +1,188 @@
...
+        'DEATH_NOTIFICATION',
+        's3://user:password@localhost', true);
+
+insert into consumer_subscription (consumer_id, event_type_id, push_uri)
+values ((select id from event_consumer WHERE consumer_name = 'S3 Consumer'),
+        'LIFE_EVENT',
+        's3://user:password@localhost');
+
+insert into consumer_subscription (consumer_id, event_type_id, push_uri)
+values ((select id from event_consumer WHERE consumer_name = 'Webhook Consumer'),
```
olevett, line 186: I'm mildly uncomfortable with this set of data potentially ending up in production/non-local dev locations, can we split it out as a dev script/something generated through infra rather than through the migrations?

src/main/kotlin/uk/gov/gdx/datashare/repository/EventData.kt
```diff
@@ -9,13 +9,12 @@ import java.time.LocalDateTime
 data class EventData(
   @Id
   val eventId: String,
-  val eventType: String,
-  val dataProvider: String,
-  val datasetType: String,
+  val eventTypeId: String,
```
olevett, line 12: Could we use something like
```@Enumerated(EnumType.STRING)```
to shift the mapping to strings down to the framework, rather than something we need to worry about directly?

**Comment by olevett** [comment]
Couple of minor comments, the thought I'd like to flesh out is around whether this ties us to having the same input and output event types - I wonder if we should be decoupling them, so there's an explicit "we're generating this event data" rather than tying us to any upstream source.

File: src/main/kotlin/uk/gov/gdx/datashare/service/DataProcessor.kt [hunk 5]
```diff
@@ -27,26 +27,26 @@ class DataProcessor(
...
     runBlocking {
       val dataProcessorMessage: DataProcessorMessage = mapper.readValue(message, DataProcessorMessage::class.java)
-      log.info("Received event [{}] from [{}]", dataProcessorMessage.eventType, dataProcessorMessage.provider)
+      log.info("Received event [{}] from [{}]", dataProcessorMessage.eventTypeId, dataProcessorMessage.publisher)
 
       // audit the event
       auditService.sendMessage(
         auditType = AuditType.EVENT_OCCURRED,
-        id = dataProcessorMessage.eventType.toString(),
+        id = dataProcessorMessage.eventTypeId.toString(),
```
oskarwilliams378, line 35: Do we need these .toString() now?

src/main/kotlin/uk/gov/gdx/datashare/repository/EventPublisher.kt
```diff
@@ -0,0 +1,24 @@
...
+import org.springframework.data.annotation.Id
+import org.springframework.data.annotation.Transient
+import org.springframework.data.domain.Persistable
+import java.time.LocalDateTime
+
+data class EventPublisher(
+  @Id
+  val id: Long,
+  val publisherName: String,
+  val whenCreated: LocalDateTime? = null,
```
oskarwilliams378, line 13: should these whenCreated fields be nullable? I feel like they should always have a value

src/main/kotlin/uk/gov/gdx/datashare/service/EventPollService.kt
```diff
@@ -1,78 +1,55 @@
...
-    // save the time of the last record
-    val lastPollEventTime = events.lastOrNull()?.eventTime ?: lastTime
-    if (dataConsumer.lastPollEventTime == null || lastPollEventTime.isAfter(dataConsumer.lastPollEventTime)) {
-      dataConsumerRepository.save(dataConsumer.copy(lastPollEventTime = lastPollEventTime))
-    }
-
-    log.info("Client: {}: Retrieved {} events between {} and {} for event types {}", dataConsumer.clientName, events.count(), beginTime, lastTime, eventTypesToPoll)
-    return events
+    return consumerSubscriptionRepository.findAllByPollerClientId(authenticationFacade.getUsername())
+      .filter { eventTypes.isNullOrEmpty() || it.eventTypeId in eventTypes }
```
oskarwilliams378, line 37: So are we saying no event types provided means you get all? I think that may be wrong.

src/main/kotlin/uk/gov/gdx/datashare/resource/EventPlatform.kt
```diff
@@ -36,8 +36,8 @@ class EventPlatform(
     ]
   )
   suspend fun getEvents(
-    @Schema(description = "Event Types required, if none supplied it will be the allowed types for this client", required = false)
-    @RequestParam(name = "eventType") eventTypes: List<EventType> = listOf(),
+    @Schema(description = "Event Types required, if none supplied it will be the allowed types for this client", required = false, allowableValues = [ "DEATH_NOTIFICATION", "LIFE_EVENT"])
```
oskarwilliams378, line 39: This says required, and then `required = false`

src/main/kotlin/uk/gov/gdx/datashare/service/EventPollService.kt
```diff
@@ -1,78 +1,55 @@
...
+      .flatMapMerge { sub ->
+        val beginTime = fromTime ?: sub.lastPollEventTime ?: now.minusDays(1)
+        val events = eventDataRepository.findAllByEventType(sub.eventTypeId, beginTime, lastTime)
+
+        val lastPollEventTime = events.lastOrNull()?.whenCreated ?: lastTime
```
oskarwilliams378, line 42: I think we should say the last poll time is the last poll time, not say its earlier than it was

src/main/kotlin/uk/gov/gdx/datashare/service/LegacyAdaptorInbound.kt
```diff
@@ -38,38 +33,6 @@ class LegacyAdaptorInbound(
     val log: Logger = LoggerFactory.getLogger(this::class.java)
   }
 
-  @Scheduled(fixedRate = 1, timeUnit = TimeUnit.MINUTES)
```
oskarwilliams378, line 41: Do we want to remove this? We can leave this in as an example of how to connect to an FTP server if we comment out the schedule and add a comment explaining what it is

oskarwilliams378 (changes requested):
I think I agree with what Olly said, having separate input and output event types will I think help us quite a lot going forward.

src/main/kotlin/uk/gov/gdx/datashare/repository/PublisherSubscriptionRepository.kt
Mjwillis: Done

src/main/resources/db/migration/V1_4__new_schema.sql
Mjwillis: Agreed

src/main/kotlin/uk/gov/gdx/datashare/repository/EventData.kt
Mjwillis: This isn't JPA so can't use this at the moment.  Reactive does work with JPA - However we can change this in future

src/main/kotlin/uk/gov/gdx/datashare/service/EventPollService.kt
Mjwillis: No you just get the types you have subscribed to

src/main/kotlin/uk/gov/gdx/datashare/service/EventPollService.kt
Mjwillis: yes ok

src/main/kotlin/uk/gov/gdx/datashare/service/DataProcessor.kt
Mjwillis: Done

src/main/kotlin/uk/gov/gdx/datashare/resource/EventPlatform.kt
Mjwillis: Done

src/main/kotlin/uk/gov/gdx/datashare/service/LegacyAdaptorInbound.kt
Mjwillis: I think remove

Issue #97, opened 2022-12-20
Fix implementation of legacy inbound adaptor S3 client


oskarwilliams378:
```
java.lang.NoSuchMethodError: 'okhttp3.Request$Builder okhttp3.Request$Builder.tag(kotlin.reflect.KClass, java.lang.Object)'
	at aws.smithy.kotlin.runtime.http.engine.okhttp.OkHttpUtilsKt.toOkHttpRequest(OkHttpUtils.kt:46)
	at aws.smithy.kotlin.runtime.http.engine.okhttp.OkHttpEngine.roundTrip(OkHttpEngine.kt:40)
	at aws.smithy.kotlin.runtime.http.SdkHttpClient$executeWithCallContext$2.invokeSuspend(SdkHttpClient.kt:54)
	at aws.smithy.kotlin.runtime.http.SdkHttpClient$executeWithCallContext$2.invoke(SdkHttpClient.kt)
	at aws.smithy.kotlin.runtime.http.SdkHttpClient$executeWithCallContext$2.invoke(SdkHttpClient.kt)
	at kotlinx.coroutines.intrinsics.UndispatchedKt.startUndispatchedOrReturn(Undispatched.kt:89)
	at kotlinx.coroutines.BuildersKt__Builders_commonKt.withContext(Builders.common.kt:169)
	at kotlinx.coroutines.BuildersKt.withContext(Unknown Source)
	at aws.smithy.kotlin.runtime.http.SdkHttpClient.executeWithCallContext(SdkHttpClient.kt:53)
	at aws.smithy.kotlin.runtime.http.SdkHttpClient.call(SdkHttpClient.kt:46)
	at aws.smithy.kotlin.runtime.http.SdkHttpClient.call(SdkHttpClient.kt:38)
	at aws.smithy.kotlin.runtime.http.operation.SdkOperationExecutionKt.httpTraceMiddleware(SdkOperationExecution.kt:179)
	at aws.smithy.kotlin.runtime.http.operation.SdkOperationExecutionKt.access$httpTraceMiddleware(SdkOperationExecution.kt:1)
	at aws.smithy.kotlin.runtime.http.operation.SdkOperationExecutionKt$decorate$1.invoke(SdkOperationExecution.kt:77)
	at aws.smithy.kotlin.runtime.http.operation.SdkOperationExecutionKt$decorate$1.invoke(SdkOperationExecution.kt:77)
[...]

PR #98, opened 2022-12-20
docs: Add documentation for WIP s3 client work

oskarwilliams378:
S3 client currently is non function, see #97 
For the moment we are leaving this as it is non blocking


Mjwillis (approved):
LGTM

Issue #72, opened 2022-12-15
Properly configure HMRC url


olevett:
We currently have mock value for `API_BASE_URL_HMRC`, we should figure out a better approach for this.

As we consume from an API directly, the simplest option for POC would be to stub data within the service itself, and not call out to an external endpoint, so we should
- remove the config value
- generate a NinoRecord programatically from `HmrcApiService`

PR #100, opened 2022-12-20
Properly configure hmrc url

CarlyG55:
https://github.com/alphagov/gdx-data-share-poc/issues/72

We currently have mock value for `API_BASE_URL_HMRC`, we should figure out a better approach for this.

As we consume from an API directly, the simplest option for POC would be to stub data within the service itself, and not call out to an external endpoint, so we should

remove the config value
generate a NinoRecord programatically from `HmrcApiService`

### DIFF
--- a/src/main/kotlin/uk/gov/gdx/datashare/service/HmrcApiService.kt
+++ b/src/main/kotlin/uk/gov/gdx/datashare/service/HmrcApiService.kt
@@ -1,22 +1,17 @@
 package uk.gov.gdx.datashare.service
 
 import org.springframework.stereotype.Service
-import org.springframework.web.reactive.function.client.WebClient
-import org.springframework.web.reactive.function.client.bodyToMono
-import reactor.core.publisher.Mono
 import java.time.LocalDate
+import java.time.format.DateTimeFormatter
 
 @Service
-class HmrcApiService(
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/gdx/datashare/service/HmrcApiService.kt
```diff
@@ -1,22 +1,17 @@
...
-  suspend fun findNiNoByNameAndDob(surname: String, firstname: String, dob: LocalDate): Mono<NinoRecord> {
-    return hmrcApiWebClient.get()
-      .uri("/hmrc/surname/$surname/firstname/$firstname/dob/$dob")
-      .retrieve()
-      .bodyToMono<NinoRecord>()
-      .onErrorResume { Mono.empty() }
+  fun generateNiNoFromNameAndDob(surname: String, firstname: String, dob: LocalDate): NinoRecord {
+    val id = (firstname + surname + dob.toString()).hashCode().toLong()
+    val niNumber = surname.substring(0, 1) + firstname.substring(0, 1) + dob.format(DateTimeFormatter.BASIC_ISO_DATE)
+      .substring(2) + surname.substring(surname.length - 1)
```
CarlyG55, line 13: Happy for opinions on something nicer here!

PR #105, opened 2022-12-21
Separate ingress and egress events

oskarwilliams378:
Splits ingress and egress events and adds mapping with 0 change between them

### DIFF
--- a/src/main/kotlin/uk/gov/gdx/datashare/repository/IngressEventType.kt
+++ b/src/main/kotlin/uk/gov/gdx/datashare/repository/IngressEventType.kt
@@ -6,27 +6,22 @@ import org.springframework.data.annotation.Transient
 import org.springframework.data.domain.Persistable
 import org.springframework.data.relational.core.mapping.Column
 import java.time.LocalDateTime
-import java.util.*
 
-data class EventData(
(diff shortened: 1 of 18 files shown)


src/main/kotlin/uk/gov/gdx/datashare/repository/IngressEventType.kt
```diff
@@ -7,11 +7,11 @@ import org.springframework.data.domain.Persistable
 import org.springframework.data.relational.core.mapping.Column
 import java.time.LocalDateTime
 
-data class EventType(
+data class IngressEventType(
 
   @Id
   @Column("id")
-  val eventId: String,
+  val eventTypeId: String,
```
Mjwillis, line 14: Notice you are leaving this as is - iD not changing to UUID

Mjwillis (approved):
LGTM

src/main/kotlin/uk/gov/gdx/datashare/repository/IngressEventType.kt
oskarwilliams378: Yes as this is the full string of the event, like DEATH_NOTIFICATION

PR #106, opened 2022-12-21
Create metrics dashboard

CarlyG55:
Once the application is deployed and running, we'll want to be able to monitor things like queue throughput, availability, latency etc.

CloudWatch has enough tooling to be able to do what we need, and Spring Boot has built in metrics (through Micrometer) which can integrate with CloudWatch.

We should

configure Spring Boot to feed metrics to Cloud Watch - see [Publishing Metrics from Spring Boot to Amazon CloudWatch](https://reflectoring.io/spring-aws-cloudwatch/) for some background, we want to use Micrometer to have a minimal amount of configuration and management
set up a minimal dashboard through terraform
configure at least one metric to ensure it all feeds through - lets use /event-data-receiver for ingested events, incrementing a counter for each call to the API
Later, we'll need to consider what the most important metrics are and what we need to worry about

### DIFF
--- a/terraform/modules/data-share-service/cloudwatch.tf
+++ b/terraform/modules/data-share-service/cloudwatch.tf
@@ -11,3 +11,27 @@ resource "aws_cloudwatch_log_group" "lb_sg_update" {
 
   kms_key_id = aws_kms_key.log_key.arn
 }
+
+resource "aws_cloudwatch_dashboard" "metrics_dashboard" {
+  dashboard_name = "metrics-dashboard"
+  dashboard_body = <<EOF
+{
+  "widgets": [
+    {
+      "type": "metric",
+      "properties": {
+        "metrics": [
+          [
+            "gdxApp",
+            "API_CALLS.IngestedEvents.count"
+          ]
+        ],
+        "period": 300,
+        "region": ${var.region},
+        "title": "API calls"
+      }
+    }
+  ]
+}
+EOF
+}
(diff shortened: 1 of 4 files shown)


Mjwillis (approved):
LGTM

terraform/modules/data-share-service/cloudwatch.tf
```diff
@@ -10,4 +10,28 @@ resource "aws_cloudwatch_log_group" "lb_sg_update" {
...
+          ]
+        ],
+        "period": 300,
+        "region": ${var.region},
+        "title": "API calls"
+      }
+    }
+  ]
+}
+EOF
```
oskarwilliams378, line 36: I don't think EOF actually will play nice here, my suggestion is use jsonencode

```suggestion
  dashboard_body = jsonencode(
{
  "widgets": [
    {
      "type": "metric",
      "properties": {
        "metrics": [
          [
            "gdxApp",
            "API_CALLS.IngestedEvents.count"
          ]
        ],
        "period": 300,
        "region": var.region,
        "title": "API calls"
      }
    }
  ]
})
```

oskarwilliams378 (changes requested):

terraform/modules/data-share-service/cloudwatch.tf
CarlyG55: Yeah this was the way the docs did it, but I can't see why else it's not working at that point, so will change it

oskarwilliams378 (approved):

PR #110, opened 2022-12-22
Add more api metrics

CarlyG55:


src/main/kotlin/uk/gov/gdx/datashare/resource/EventPlatform.kt
```diff
@@ -22,9 +24,10 @@ import java.util.UUID
 @PreAuthorize("hasAnyAuthority('SCOPE_events/poll')")
 @Validated
 class EventPlatform(
-  private val eventPollService: EventPollService
+  private val eventPollService: EventPollService,
+  meterRegistry: MeterRegistry,
 ) {
-
+  private val callsToPollCounter: Counter = meterRegistry.counter("API_CALLS.CallsToPoll")
```
oskarwilliams378, line 30: new line after this please :)

src/main/kotlin/uk/gov/gdx/datashare/service/HmrcApiService.kt
```diff
@@ -1,19 +1,25 @@
...
 import java.time.format.DateTimeFormatter
 import java.util.*
 
 @Service
-class HmrcApiService() {
-
+class HmrcApiService(meterRegistry: MeterRegistry) {
+  private val callsToHmrcCounter: Counter = meterRegistry.counter("API_CALLS.CallsToHmrc")
+  private val responsesFromHmrcCounter: Counter = meterRegistry.counter("API_RESPONSES.ResponsesFromHmrc")
   fun generateNiNoFromNameAndDob(surname: String, firstname: String, dob: LocalDate): NinoRecord {
```
oskarwilliams378, line 14: Could we add a function here called something like `getNiNo` which takes the same arguments, this would then be the public exposure of this method and would put the calls around it , something like the below, as that would mean there is a very obvious place where we would put in an actual call to an API

```
  private val callsToHmrcCounter: Counter = meterRegistry.counter("API_CALLS.CallsToHmrc")
  private val responsesFromHmrcCounter: Counter = meterRegistry.counter("API_RESPONSES.ResponsesFromHmrc")

  fun getNiNo(surname: String, firstname: String, dob: LocalDate): NinoRecord {
    callsToHmrcCounter.increment()
    val nino = generateNiNoFromNameAndDob(surname, firstname, dob)
    responsesFromHmrcCounter.increment()
    return nino
  }
  
  private fun generateNiNoFromNameAndDob(surname: String, firstname: String, dob: LocalDate): NinoRecord {
    val uniqueIdentifier = (firstname + surname + dob.toString()).hashCode().toLong()
    val id = UUID(uniqueIdentifier, 0)
    val niNumber = surname.substring(0, 1) + firstname.substring(0, 1) + dob.format(DateTimeFormatter.BASIC_ISO_DATE)
      .substring(2) + surname.substring(surname.length - 1)
    return NinoRecord(id = id, ni_number = niNumber)
  }
```

src/main/kotlin/uk/gov/gdx/datashare/service/LevApiService.kt
```diff
@@ -11,15 +13,20 @@ import java.time.LocalDate
 
 @Service
 class LevApiService(
-  private val levApiWebClient: WebClient
+  private val levApiWebClient: WebClient,
+  meterRegistry: MeterRegistry
 ) {
-
+  private val callsToLevCounter: Counter = meterRegistry.counter("API_CALLS.CallsToLev")
+  private val responsesFromLevCounter: Counter = meterRegistry.counter("API_RESPONSES.ResponsesFromLev")
```
oskarwilliams378, line 20: new line after this please

terraform/modules/data-share-service/cloudwatch.tf
```diff
@@ -23,6 +23,30 @@ resource "aws_cloudwatch_dashboard" "metrics_dashboard" {
             [
               "gdxApp",
               "API_CALLS.IngestedEvents.count"
+            ],
+            [
+              ".",
+              "API_CALLS.CallsToLev.count"
```
oskarwilliams378, line 29: it may be worth explicitly setting the colour :)

oskarwilliams378 (changes requested):

terraform/modules/data-share-service/cloudwatch.tf
```diff
@@ -23,6 +23,30 @@ resource "aws_cloudwatch_dashboard" "metrics_dashboard" {
...
+            [
+              ".",
+              "API_CALLS.CallsToPoll.count"
+            ],
+            [
+              ".",
+              "API_CALLS.CallsToEnrich.count"
             ]
           ],
           "period" : 300,
```
oskarwilliams378, line 52: We should also add this here as currently it is doing averages
    "stat": "Sum"

oskarwilliams378 (changes requested):

oskarwilliams378 (approved):

PR #126, opened 2023-01-03
feat: Add base data controller

oskarwilliams378:


src/main/kotlin/uk/gov/gdx/datashare/controller/BaseDataController.kt
```diff
@@ -0,0 +1,303 @@
...
+    summary = "Get Ingress Events",
+    description = "Need scope of pubsub/maintain",
+    responses = [
+      ApiResponse(
+        responseCode = "200",
+        description = "Ingress Event"
+      )
+    ]
+  )
+  suspend fun getIngressEvents() = ingressEventDataRepository.findAll().toList()
```
Mjwillis, line 158: Do you need the toList?  Return a flow instead?

Mjwillis (commented):
Flow comment

src/main/kotlin/uk/gov/gdx/datashare/controller/BaseDataController.kt
oskarwilliams378: removed the to list :)

src/main/kotlin/uk/gov/gdx/datashare/controller/AdminController.kt
```diff
@@ -1,4 +1,4 @@
-package uk.gov.gdx.datashare.resource
```
olevett, line 1: I think I'd ideally have done this rename as a completely separate commit to main - it makes this a little more noisy than would be ideal

src/main/kotlin/uk/gov/gdx/datashare/controller/BaseDataController.kt
```diff
@@ -0,0 +1,303 @@
...
+      )
+    ]
+  )
+  suspend fun deletePublisherSubscription(
+    @Schema(description = "Publisher Subscription ID", required = true)
+    @PathVariable id: UUID,
+  ) = publisherSubscriptionRepository.deleteById(id)
+
+  //The below endpoints don't exist through any other means, so have added them here
+  @PostMapping("/ingressType")
```
olevett, line 264: I think we need to make sure we've fully aligned around "ingress", "publisher" etc - but that's out of this scope I think

olevett (approved):

src/main/kotlin/uk/gov/gdx/datashare/controller/AdminController.kt
oskarwilliams378: splitting

src/main/kotlin/uk/gov/gdx/datashare/controller/BaseDataController.kt
oskarwilliams378: I totally agree, this is just to keep the mapping consistent atm :)

PR #146, opened 2023-01-04
refactor: Optimise imports

oskarwilliams378:


src/main/kotlin/uk/gov/gdx/datashare/config/AuthenticationFacade.kt
```diff
@@ -7,7 +7,7 @@ import org.springframework.security.core.GrantedAuthority
 import org.springframework.security.core.context.ReactiveSecurityContextHolder
 import org.springframework.security.oauth2.jwt.Jwt
 import org.springframework.stereotype.Component
-import java.util.Arrays
+import java.util.*
```
CarlyG55, line 10: I think ktlint did the reverse of these (for all the `java.util.*`'s)

PR #149, opened 2023-01-04
Fix S3 polling

Mjwillis:
Using the aws-java-sdk instead for handling S3 - Also used a simple Apache CSV to parse the CSV
Will run locally with `aws --endpoint-url=http://localhost:4566 s3 cp death1.csv s3://ingress/` as buckets are created on the `dev` profile
A Sample CSV is 
```civ
"123456789","SMITH","Joan Narcissus Ouroboros","2008-08-08","2008-08-08","male","888 Death House, 8 Death lane, Deadington, Deadshire"
```

Haven't done tests yet and also need to make handle quotes etc on the parsing 

We should also get the Egress output to an S3 working too


src/main/kotlin/uk/gov/gdx/datashare/service/LegacyAdaptorInbound.kt
```diff
@@ -41,71 +42,80 @@ class LegacyAdaptorInbound(
...
+          log.info("Found file for processing ${it.key}")
+        }.objectContent
+        .let {
+          getCSV(it).forEach { csvRecord ->
+            with(csvRecord) {
+              postDataToReceiver(
+                EventToPublish(
+                  id = certificateId(),
+                  eventType = "DEATH_NOTIFICATION",
+                  eventDetails = "${lastName()},${firstName()},${dateOfBirth()},${dateOfDeath()},${gender()},\"${address()}\"" // TODO: decide whether to format etc
```
olevett, line 73: I think I'm _expecting_ this to be JSON-like, but I don't think we do much with it so it probably doesn't matter right now

src/main/kotlin/uk/gov/gdx/datashare/service/LegacyAdaptorInbound.kt
```diff
@@ -41,71 +42,80 @@ class LegacyAdaptorInbound(
...
-              processBucketObject(s3, it)
+              processBucketObject(it)
             }
           }
-        }
       }
-    } catch (e: NoSuchMethodError) {
-      log.error("Failed to connect to S3")
-    } catch (e: Exception) {
+    } catch (e: AmazonClientException) {
```
olevett, line 53: I'm not totally sure what the try/catch here gets us if we're just logging it out - I think an uncaught exception in a scheduled task will get logged out fairly usefully for us anyway

PR #153, opened 2023-01-05
chore: Remove HMPPS SQS dependencies

olevett:
This is a bit of a thinking out loud more than an absolute must, but from the problems we had with ShedLock, I'd be keen to reduce the number of dependencies overall/have control over them, and particularly, have control over things like spring versions.

This is a lift of the bits of the SQS/SNS management that are required for our code to function and no more, and it's not massively complex code in the end either so I think the maintenance burden for us is quite low. I also think there are bits of the HMPPS code that are a useful general solution, but not something I think we really want in their current form (the DLQ retrying stuff for example, I think we want different control over to suit our use cases), so bringing that in to then be editable by us easily seems reasonable.

### DIFF
--- a/src/main/kotlin/uk/gov/gdx/datashare/queue/AmazonSnsFactory.kt
+++ b/src/main/kotlin/uk/gov/gdx/datashare/queue/AmazonSnsFactory.kt
@@ -1,0 +1,58 @@
+package uk.gov.gdx.datashare.queue
+
+import com.amazonaws.auth.AWSStaticCredentialsProvider
+import com.amazonaws.auth.BasicAWSCredentials
+import com.amazonaws.client.builder.AwsClientBuilder
+import com.amazonaws.services.sns.AmazonSNS
+import com.amazonaws.services.sns.AmazonSNSAsyncClientBuilder
+import com.amazonaws.services.sns.AmazonSNSClientBuilder
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+class AmazonSnsFactory {
+  companion object {
+    val log: Logger = LoggerFactory.getLogger(this::class.java)
+  }
+
+  fun awsSnsClient(topicId: String, accessKeyId: String, secretAccessKey: String, region: String, asyncClient: Boolean): AmazonSNS =
+    when (asyncClient) {
+      false -> awsAmazonSNS(accessKeyId, secretAccessKey, region)
+      true -> awsAmazonSNSAsync(accessKeyId, secretAccessKey, region)
+    }.also { log.info("Created an AWS SNS client for topicId=$topicId") }
+
+  fun localstackSnsClient(topicId: String, localstackUrl: String, region: String, asyncClient: Boolean): AmazonSNS =
+    when (asyncClient) {
+      false -> localstackAmazonSNS(localstackUrl, region)
+      true -> localstackAmazonSNSAsync(localstackUrl, region)
+    }.also { log.info("Created a LocalStack SNS client for topicId=$topicId") }
+
+  private fun awsAmazonSNS(accessKeyId: String, secretAccessKey: String, region: String) =
+    BasicAWSCredentials(accessKeyId, secretAccessKey)
+      .let { credentials ->
+        AmazonSNSClientBuilder.standard()
+          .withCredentials(AWSStaticCredentialsProvider(credentials))
+          .withRegion(region)
+          .build()
+      }
+
+  private fun awsAmazonSNSAsync(accessKeyId: String, secretAccessKey: String, region: String) =
+    BasicAWSCredentials(accessKeyId, secretAccessKey)
+      .let { credentials ->
+        AmazonSNSAsyncClientBuilder.standard()
+          .withCredentials(AWSStaticCredentialsProvider(credentials))
+          .withRegion(region)
+          .build()
+      }
+
+  private fun localstackAmazonSNS(localstackUrl: String, region: String) =
+    AmazonSNSClientBuilder.standard()
+      .withEndpointConfiguration(AwsClientBuilder.EndpointConfiguration(localstackUrl, region))
+      .withCredentials(AWSStaticCredentialsProvider(BasicAWSCredentials("any", "any"))) // LocalStack doesn't work with Anonymous credentials when dealing with topics but doesn't care what the credential values are.
+      .build()
+
+  private fun localstackAmazonSNSAsync(localstackUrl: String, region: String) =
+    AmazonSNSAsyncClientBuilder.standard()
+      .withEndpointConfiguration(AwsClientBuilder.EndpointConfiguration(localstackUrl, region))
+      .withCredentials(AWSStaticCredentialsProvider(BasicAWSCredentials("any", "any"))) // LocalStack doesn't work with Anonymous credentials when dealing with topics but doesn't care what the credential values are.
+      .build()
+}
--- a/src/main/kotlin/uk/gov/gdx/datashare/queue/AmazonSnsFactory.kt
+++ b/src/main/kotlin/uk/gov/gdx/datashare/queue/AmazonSnsFactory.kt
@@ -4,7 +4,6 @@ import com.amazonaws.auth.AWSStaticCredentialsProvider
 import com.amazonaws.auth.BasicAWSCredentials
 import com.amazonaws.client.builder.AwsClientBuilder
 import com.amazonaws.services.sns.AmazonSNS
-import com.amazonaws.services.sns.AmazonSNSAsyncClientBuilder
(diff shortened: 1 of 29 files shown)


Mjwillis (approved):
Lgtm

src/test/kotlin/uk/gov/gdx/datashare/integration/SqsIntegrationTestBase.kt
```diff
@@ -10,11 +10,12 @@ import org.springframework.test.context.ActiveProfiles
...
-import uk.gov.justice.hmpps.sqs.HmppsQueueService
-import uk.gov.justice.hmpps.sqs.HmppsSqsProperties
-import uk.gov.justice.hmpps.sqs.MissingQueueException
-import uk.gov.justice.hmpps.sqs.MissingTopicException
+import uk.gov.gdx.datashare.queue.HmppsQueue
+import uk.gov.gdx.datashare.queue.HmppsQueueService
+import uk.gov.gdx.datashare.queue.HmppsSqsProperties
+import uk.gov.gdx.datashare.queue.MissingQueueException
+import uk.gov.gdx.datashare.queue.MissingTopicException
+
```
oskarwilliams378, line 18: extra new line

src/main/kotlin/uk/gov/gdx/datashare/queue/HmppsQueue.kt
```diff
@@ -0,0 +1,14 @@
+package uk.gov.gdx.datashare.queue
+
+import com.amazonaws.services.sqs.AmazonSQS
+
+class HmppsQueue(
```
oskarwilliams378, line 5: Have you considered a big ol' rename of everything HMPPS to remove that?

oskarwilliams378 (changes requested):

src/main/kotlin/uk/gov/gdx/datashare/queue/AmazonSnsFactory.kt
```diff
@@ -0,0 +1,38 @@
...
+import com.amazonaws.services.sns.AmazonSNSClientBuilder
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+class AmazonSnsFactory {
+  companion object {
+    val log: Logger = LoggerFactory.getLogger(this::class.java)
+  }
+
+  fun awsSnsClient(topicId: String, accessKeyId: String, secretAccessKey: String, region: String): AmazonSNS =
```
ethanmills, line 16: Why does this take a topicId? Is it just for logging?

src/main/kotlin/uk/gov/gdx/datashare/queue/AmazonSqsFactory.kt
```diff
@@ -0,0 +1,44 @@
...
+import com.amazonaws.services.sqs.AmazonSQSClientBuilder
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+class AmazonSqsFactory {
+  companion object {
+    val log: Logger = LoggerFactory.getLogger(this::class.java)
+  }
+
+  fun awsSqsClient(queueId: String, queueName: String, accessKeyId: String, secretAccessKey: String, region: String): AmazonSQS =
```
ethanmills, line 17: Ditto for these params

src/main/kotlin/uk/gov/gdx/datashare/queue/HmppsQueueFactory.kt
```diff
@@ -0,0 +1,140 @@
...
+  companion object {
+    val log: Logger = LoggerFactory.getLogger(this::class.java)
+  }
+
+  fun createHmppsQueues(hmppsSqsProperties: HmppsSqsProperties, hmppsTopics: List<HmppsTopic> = listOf()) =
+    hmppsSqsProperties.queues
+      .map { (queueId, queueConfig) ->
+        val sqsDlqClient = getOrDefaultSqsDlqClient(queueId, queueConfig, hmppsSqsProperties)
+        val sqsClient = getOrDefaultSqsClient(queueId, queueConfig, hmppsSqsProperties, sqsDlqClient)
+          .also { subscribeToLocalStackTopic(hmppsSqsProperties, queueConfig, hmppsTopics) }
```
ethanmills, line 29: Perhaps just a convention I've not seen before, but I'm a little surprised by all these uses of `.also` without reference to `it`

src/main/kotlin/uk/gov/gdx/datashare/queue/HmppsQueueFactory.kt
```diff
@@ -0,0 +1,140 @@
...
+        "aws" -> amazonSqsFactory.awsSqsClient(queueId, queueConfig.queueName, queueConfig.queueAccessKeyId, queueConfig.queueSecretAccessKey, region)
+        "localstack" ->
+          amazonSqsFactory.localStackSqsClient(queueId, queueConfig.queueName, localstackUrl, region)
+            .also { sqsClient -> createLocalStackQueue(sqsClient, sqsDlqClient, queueConfig.queueName, queueConfig.dlqName, queueConfig.dlqMaxReceiveCount) }
+
+        else -> throw IllegalStateException("Unrecognised HMPPS SQS provider $provider")
+      }
+    }
+
+  private fun createLocalStackQueue(
```
ethanmills, line 89: If you want to be really pedantic, it's a little confusing that the top function here creates a Queue object, where as this actually creates an SQS queue

src/main/kotlin/uk/gov/gdx/datashare/queue/HmppsQueueResourceAsync.kt
```diff
@@ -0,0 +1,54 @@
...
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PutMapping
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RequestParam
+import org.springframework.web.bind.annotation.RestController
+import org.springframework.web.server.ResponseStatusException
+
+/*
+ * Warning: This is a duplicate of HmppsQueueResource but with suspend functions. Therefore, this class should be kept in sync with the non-async class.
```
ethanmills, line 14: Does HmpssQueueResource exist?

src/main/kotlin/uk/gov/gdx/datashare/queue/HmppsQueueResourceAsync.kt
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * This class *should* just be a wrapper around HmppsQueueResource but that sometimes works and sometimes doesn't! So we're keeping the duplication for now.
+ */
+@RestController
+@RequestMapping("/queue-admin")
+class HmppsQueueResourceAsync(private val hmppsQueueService: HmppsQueueService) {
```
ethanmills, line 20: should this be a controller?

ethanmills (commented):
Did the library have any tests that we'd like to bring across? I think that's all that I'd want changed before merging this - the rest is just an exercise in pedantry

src/main/kotlin/uk/gov/gdx/datashare/queue/HmppsQueueResourceAsync.kt
olevett: Ah, no, I deleted that, good shout

src/main/kotlin/uk/gov/gdx/datashare/queue/HmppsQueueResourceAsync.kt
olevett: Yes

src/main/kotlin/uk/gov/gdx/datashare/queue/HmppsQueue.kt
olevett: I was considering it, but hadn't yet - I can do so now though

src/main/kotlin/uk/gov/gdx/datashare/queue/AmazonSnsFactory.kt
olevett: Yeah, not sure how helpful the logging is right now, maybe we should do a cull of it at some point

src/main/kotlin/uk/gov/gdx/datashare/queue/HmppsQueueResourceAsync.kt
olevett: I'm tempted to outright remove this anyway, but might leave that as an exercise for later

oskarwilliams378 (approved):

PR #154, opened 2023-01-05
feat: Add script that removes dev data

oskarwilliams378:


olevett:
So this is a script to run manually as and when required?

olevett (approved):

PR #161, opened 2023-01-05
chore: Upgrade Kotlin Spring/Java plugins

olevett:


olevett:
Replaces #150  and #113

PR #173, opened 2023-01-06
Further metrics

CarlyG55:
https://trello.com/c/bjLGFOYL/29-metrics-for-dashboarding

Adding these metrics first, then will add to terraform

### DIFF
--- a/src/main/kotlin/uk/gov/gdx/datashare/controller/EventsController.kt
+++ b/src/main/kotlin/uk/gov/gdx/datashare/controller/EventsController.kt
@@ -33,10 +33,12 @@ class EventsController(
   meterRegistry: MeterRegistry,
 ) {
   private val publishEventCounter: Counter = meterRegistry.counter("API_CALLS.PublishEvent")
+  private val publishEventSuccessCounter: Counter = meterRegistry.counter("SUCCESSFUL_API_CALLS.PublishEvent")
   private val getEventCounter: Counter = meterRegistry.counter("API_CALLS.GetEvent")
   private val getEventsCounter: Counter = meterRegistry.counter("API_CALLS.GetEvents")
   private val getEventsStatusCounter: Counter = meterRegistry.counter("API_CALLS.GetEventsStatus")
   private val deleteEventCounter: Counter = meterRegistry.counter("API_CALLS.DeleteEvent")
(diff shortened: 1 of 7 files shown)


oskarwilliams378 (changes requested):
Have you checked that whenCreated is the right property to check? (just to be safe)

oskarwilliams378 (approved):

src/main/kotlin/uk/gov/gdx/datashare/controller/EventsController.kt
```diff
@@ -33,10 +33,12 @@ class EventsController(
   meterRegistry: MeterRegistry,
 ) {
   private val publishEventCounter: Counter = meterRegistry.counter("API_CALLS.PublishEvent")
+  private val publishEventSuccessCounter: Counter = meterRegistry.counter("SUCCESSFUL_API_CALLS.PublishEvent")
```
olevett, line 36: What's the distinction between "SUCCESSFUL_API_CALLS" and "API_CALLS"? It looks like we might be incrementing these two in pretty much the same place? 

This might be a case that using tags but the same basic identifier might be clearer

olevett (approved):

src/main/kotlin/uk/gov/gdx/datashare/controller/EventsController.kt
CarlyG55: After chatting to @oskarwilliams378 about

> for a given time window, how many events are we receiving, e.g. how many DEATH_EVENTS did we receive in the last 5 minutes (this could conceivably be both "how many calls to our API are happening" and also "how many API calls result in something on a queue")

on the ticket, we thought it would make sense to have metrics for how many times the endpoint is called but also how many successes there are. Not sure if there was a method we explicitly though would fail

olevett: The other metric that we should have somewhere that might be relevant is HTTP status codes, how many calls to this API return 200 vs how man return a 500 or something else?

CarlyG55: What I've done is add a tag for "success", and gotten rid of the total one - so we can track successes & failures by using the tag, but if we don't use the tag we get total number of calls

olevett (approved):

PR #183, opened 2023-01-09
feat: add bastion host to allow access to RDS instances

ethanmills:


ethanmills:
Fixing TFSec and generating a keypair for SSH access

terraform/modules/data-share-service/ec2.tf
```diff
@@ -0,0 +1,75 @@
...
+    encrypted = true
+  }
+}
+
+resource "aws_key_pair" "rds_bastion_key_pair" {
+  public_key = module.tls_private_key.public_key_pem
+  key_name   = "rds-bastion-key-pair-${var.environment}"
+}
+
+#tfsec:ignore:aws-ec2-no-public-ingress-sgr
```
oskarwilliams378, line 44: Is there any sensible limit we can put on this?

terraform/modules/data-share-service/rds.tf
```diff
@@ -48,23 +47,23 @@ resource "aws_rds_cluster_instance" "db_aurora" {
 
 resource "aws_security_group" "rds_postgres_cluster" {
   name_prefix = "${var.environment}-rds-postgres-cluster-"
-  description = "For RDS cluster, inbound access from ECS only"
+  description = "For RDS cluster, inbound access from ECS, bastion host only"
```
oskarwilliams378, line 50:
```suggestion
  description = "For RDS cluster, inbound access from ECS or bastion host only"
```
Is that what this means?

terraform/modules/data-share-service/rds.tf
```diff
@@ -48,23 +47,23 @@ resource "aws_rds_cluster_instance" "db_aurora" {
...
   vpc_id      = module.vpc.vpc_id
 
   ingress {
     protocol        = "tcp"
     from_port       = 5432
     to_port         = 5432
-    security_groups = [aws_security_group.ecs_tasks.id]
-    description     = "ECS task ingress rule, allow access from ECS tasks only"
+    security_groups = [aws_security_group.ecs_tasks.id, aws_security_group.rds_bastion_host_sg.id]
+    description     = "ECS task ingress rule, allow access from ECS tasks, bastion host only"
```
oskarwilliams378, line 58: similar comment

oskarwilliams378 (changes requested):

terraform/modules/data-share-service/ec2.tf
ethanmills: We could use an IP allow list? I don't know what values we would use other than the SW IP ranges though

oskarwilliams378: for the moment this is good, but could we add a TODO

ethanmills: Created ticket

PR #190, opened 2023-01-10
WIP - Outbound adaptor functionality

Mjwillis:
- Added egress bucket
- switched to check the reading client id rather than poller ( these can actually be merged into one )
- Example of consuming off the queue (but will add the polling mechanism next)


src/main/kotlin/uk/gov/gdx/datashare/repository/EgressEventDataRepository.kt
```diff
@@ -63,6 +63,15 @@ interface EgressEventDataRepository : CoroutineCrudRepository<EgressEventData, U
   )
   suspend fun findByPollClientIdAndId(pollerClientId: String, id: UUID): EgressEventData?
 
+  @Query(
```
Mjwillis, line 66: We need to get rid of both client id columns - only need one

src/main/kotlin/uk/gov/gdx/datashare/service/DataShareTopicService.kt
```diff
@@ -24,11 +24,12 @@ class DataShareTopicService(awsQueueService: AwsQueueService, private val object
   }
   private val domainEventsTopicClient by lazy { domainEventsTopic.snsClient }
 
-  fun sendGovEvent(eventId: UUID, occurredAt: LocalDateTime, eventType: String) {
+  fun sendGovEvent(eventId: UUID, consumer: String, occurredAt: LocalDateTime, eventType: String) {
```
Mjwillis, line 27: Adding consumer as possible idea for filtering topic by consumer, (used in the outbound example)

src/main/kotlin/uk/gov/gdx/datashare/service/EventDataService.kt
```diff
@@ -56,7 +56,7 @@ class EventDataService(
     id: UUID
   ): EventNotification? {
     val clientId = authenticationFacade.getUsername()
-    val event = egressEventDataRepository.findByPollClientIdAndId(clientId, id)
+    val event = egressEventDataRepository.findByCallbackClientIdAndId(clientId, id)
```
Mjwillis, line 59: Combine into one column in future commit

src/main/resources/application-localstack.yml
```diff
@@ -10,12 +10,12 @@ sqs:
       queueName: legacy-adaptor-event-queue
       dlqName: legacy-adaptor-event-dlq
       subscribeTopicId: event
-      subscribeFilter: '{"eventType":[ "LIFE_EVENT", "DEATH_NOTIFICATION" ] }'
+      subscribeFilter: '{"consumer":[ "Internal Adaptor" ] }'
```
Mjwillis, line 13: Queue only goes to one consumer

PR #220, opened 2023-01-12
feat: Add LEV DB instead of mock usage

oskarwilliams378:


terraform/modules/lev_api/main.tf
```diff
@@ -20,6 +20,31 @@ data "aws_iam_policy_document" "lev_api_ecr_role_assume_policy" {
   }
 }
 
+resource "aws_security_group" "lev_api" {
```
olevett, line 23: TFsec won't like this

terraform/modules/lev_api/main.tf
```diff
@@ -43,9 +68,13 @@ resource "aws_apprunner_service" "lev_api" {
...
-          LISTEN_HOST = "0.0.0.0"
-          LISTEN_PORT = "8000"
-          MOCK        = "true"
+          LISTEN_HOST       = "0.0.0.0"
+          LISTEN_PORT       = "8000"
+          POSTGRES_USER     = random_string.rds_username.result
+          POSTGRES_PASSWORD = random_password.rds_password.result
+          POSTGRES_HOST     = aws_rds_cluster.lev_rds_postgres_cluster.endpoint
+          POSTGRES_DB       = aws_rds_cluster.lev_rds_postgres_cluster.database_name
+          POSTGRES_SSL      = "false"
```
olevett, line 77: Is there a reason we're not using SSL?

olevett (approved):

terraform/modules/lev_api/main.tf
oskarwilliams378: I've made it shh

PR #221, opened 2023-01-12
feat: Add 2 min wait time on code deploy to let instances start up

oskarwilliams378:


ethanmills (approved):
Looks like a good quick fix 🙂

PR #222, opened 2023-01-12
fix: Give LEN lambda permission to send data to x-ray

ethanmills:


terraform/modules/len/main.tf
```diff
@@ -42,9 +42,14 @@ resource "aws_iam_role" "len" {
...
+resource "aws_iam_role_policy_attachment" "len_xray_access" {
+  role       = aws_iam_role.len.name
+  policy_arn = "arn:aws:iam::aws:policy/AWSXRayDaemonWriteAccess"
+}
+
 resource "aws_cloudwatch_event_rule" "schedule" {
-  name                = "schedule"
-  description         = "Schedule for Lambda Function"
+  name                = "LENLambdaSchedule"
+  description         = "Schedule for Lambda Function to send LEN events"
```
ethanmills, line 52: We're going to have a few of these and they're pretty hard to distinguish in the console. Open to better naming suggestions @oskarwilliams378

terraform/modules/len/main.tf
```diff
@@ -42,9 +42,14 @@ resource "aws_iam_role" "len" {
   assume_role_policy = data.aws_iam_policy_document.len_assume_policy.json
 }
 
+resource "aws_iam_role_policy_attachment" "len_xray_access" {
+  role       = aws_iam_role.len.name
+  policy_arn = "arn:aws:iam::aws:policy/AWSXRayDaemonWriteAccess"
```
ethanmills, line 47: This is currently
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "xray:PutTraceSegments",
                "xray:PutTelemetryRecords",
                "xray:GetSamplingRules",
                "xray:GetSamplingTargets",
                "xray:GetSamplingStatisticSummaries"
            ],
            "Resource": [
                "*"
            ]
        }
    ]
}
```
per [here](https://docs.aws.amazon.com/lambda/latest/dg/lambda-intro-execution-role.html#:~:text=Select%20the%20AWS%20managed%20policies%20AWSLambdaBasicExecutionRole%20and%20AWSXRayDaemonWriteAccess.)

ethanmills:
I'm a little surprised that the log permissions we've granted are sufficient. The policy we have in place is 
```json
{
    effect = "Allow"
    actions = [
      "logs:CreateLogStream",
      "logs:PutLogEvents",
      "logs:AssociateKmsKey"
    ]
    resources = ["arn:aws:logs:${var.region}:${data.aws_caller_identity.current.account_id}:log-group:*"]
  }
```
where as the [AWSLambdaBasicExecutionRole](https://us-east-1.console.aws.amazon.com/iam/home#/policies/arn%3Aaws%3Aiam%3A%3Aaws%3Apolicy%2Fservice-role%2FAWSLambdaBasicExecutionRole) has 
```json 
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "logs:CreateLogGroup",
                "logs:CreateLogStream",
                "logs:PutLogEvents"
            ],
            "Resource": "*"
        }
    ]
}
```
which includes ` "logs:CreateLogGroup"`

terraform/modules/data-share-service/codedeploy.tf
```diff
@@ -16,7 +16,7 @@ resource "aws_codedeploy_deployment_group" "gdx_data_share_poc" {
 
   blue_green_deployment_config {
     deployment_ready_option {
-      action_on_timeout = "CONTINUE_DEPLOYMENT"
+      action_on_timeout    = "CONTINUE_DEPLOYMENT"
```
ethanmills, line 19: Unrelated but made the format checker unhappy

terraform/modules/len/main.tf
olevett: Can we include at least the environment name in the description if we've got it?

olevett (approved):

terraform/modules/data-share-service/codedeploy.tf
oskarwilliams378: yes thanks, I am blind

PR #227, opened 2023-01-13
housekeeping:  Ktlint

olevett:
Vaguely related to the terraform stuff, this is some thoughts on Kotlin linting

We've got inconsistency in the trailing comma rule, which may or may not cause a fight - I'm easy either way...

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -13,6 +13,10 @@ ij_java_class_count_to_use_import_on_demand = 99
 ij_java_names_count_to_use_import_on_demand = 99
 ij_java_use_single_class_imports = true
 ij_java_insert_inner_class_imports = true
+ij_kotlin_allow_trailing_comma = true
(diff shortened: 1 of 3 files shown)


oskarwilliams378 (approved):

src/main/kotlin/uk/gov/gdx/datashare/config/ApiExceptionHandler.kt
```diff
@@ -36,8 +36,8 @@ class ApiExceptionHandler {
         ErrorResponse(
           status = (BAD_REQUEST.value()),
           userMessage = "Validation failure: $message",
-          developerMessage = (e.message)
-        )
+          developerMessage = (e.message),
+        ),
```
ethanmills, line 40: This makes me a little sad, but not enough to kick up a fuss

ethanmills (approved):

PR #231, opened 2023-01-13
Get event lambda

ethanmills:
Get an event from the datashare API and validate the fields against the LEV API. I intend to add CW metrics and dashboarding for this, but that will come later

### DIFF
--- a/terraform/modules/consumer/lambdas/get_event.py
+++ b/terraform/modules/consumer/lambdas/get_event.py
@@ -1,0 +1,74 @@
+import json
+import logging
+import os
+from datetime import datetime
+from http.client import HTTPResponse
+from urllib import request
+
+from common import get_auth_token
+
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+
+gdx_url = os.environ["gdx_url"]
+events_url = gdx_url + "/events"
+
+auth_url = os.environ["auth_url"]
+
+client_id = os.environ["client_id"]
+client_secret = os.environ["client_secret"]
+
+lev_api_url = os.environ["lev_api_url"]
+
+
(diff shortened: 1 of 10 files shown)


terraform/modules/lambdas/common.py
```diff
@@ -0,0 +1,18 @@
+import json
```
ethanmills, line 1: I've done a bit of commonisation here. Open to being told this is a terrible idea

terraform/modules/len/lambdas/len.py
```diff
@@ -1,10 +1,12 @@
...
 import math
 import os
-from random import randint
-from urllib import request, parse
-import json
 from datetime import datetime
+from random import randint
+from urllib import request
+
+from common import get_auth_token
```
ethanmills, line 9: Sorry, this auto optimize on save

terraform/modules/consumer/lambdas/get_event.py
```diff
@@ -0,0 +1,71 @@
...
+
+gdx_url = os.environ["gdx_url"]
+events_url = gdx_url + "/events"
+
+auth_url = os.environ["auth_url"]
+
+client_id = os.environ["client_id"]
+client_secret = os.environ["client_secret"]
+
+lev_api_url = os.environ["lev_api_url"]
```
olevett, line 21: I find this surprising, I'd have assumed that our event consumption is ignorant of anything but GDX, are we doing some validation of things against LEV?

olevett: I scrolled a bit further on and found the answer. I think this is worth an explanatory note though, in that normal consumers would not need to do this.

olevett (approved):

PR #239, opened 2023-01-16
GET /events/{id} should return a 404 or similar when an event isn't found rather than a 500

Mjwillis:


src/main/kotlin/uk/gov/gdx/datashare/service/EventDataService.kt
```diff
@@ -117,9 +118,9 @@ class EventDataService(
   suspend fun deleteEvent(id: UUID) {
     val callbackClientId = authenticationFacade.getUsername()
     val egressEvent = egressEventDataRepository.findByClientIdAndId(callbackClientId, id)
-      ?: throw NotFoundException("Egress event $id not found for callback client $callbackClientId")
+      ?: throw EventNotFoundException("Egress event $id not found for callback client $callbackClientId")
     val consumerSubscription = consumerSubscriptionRepository.findByEgressEventId(id)
-      ?: throw NotFoundException("Consumer subscription not found for egress event $id")
+      ?: throw EventNotFoundException("Consumer subscription not found for egress event $id")
```
ethanmills, line 123: Is this the right type?

ethanmills (approved):

src/main/kotlin/uk/gov/gdx/datashare/service/EventDataService.kt
Mjwillis: possibly not!

PR #241, opened 2023-01-16
Add a few more tests around Not Found return status

Mjwillis:


src/test/resources/application-test.yml
```diff
@@ -43,6 +43,7 @@ spring:
     url: jdbc:postgresql://localhost:5432/datashare?sslmode=prefer
     user: datashare
     password: datashare
+    locations: classpath:db/migration, classpath:db/devonly
```
ethanmills, line 46: This change seems unrelated. Is it deliberate?

ethanmills (approved):

PR #249, opened 2023-01-17
fix: add RDS cert to truststore

ethanmills:
Seeing errors like image when connecting to database.  Not sure this is the fix, but it looks promising 
image


Mjwillis (approved):
LGTM

PR #250, opened 2023-01-17
Make /events endpoint return minimal information

Mjwillis:


src/main/kotlin/uk/gov/gdx/datashare/repository/ConsumerSubscription.kt
```diff
@@ -30,9 +30,16 @@ data class ConsumerSubscription(
   @Schema(description = "Events Type of Ingress Notification", required = true, example = "DEATH_NOTIFICATION")
   val ingressEventType: String,
   @Schema(
-    description = "CSV List of required fields to enrich the event with",
+    description = "Indicates that the specified enrichment fields will be present when a poll of events occurs",
```
olevett, line 33: Can we sensibly link this "poll of events" comment to the `/events` api, even if that's just a comment to that effect I think it would be useful

src/test/kotlin/uk/gov/gdx/datashare/service/EventDataServiceTest.kt
```diff
@@ -243,6 +243,43 @@ class EventDataServiceTest {
     }
   }
 
+  @Test
+  fun `getEvents gets thin events for client`() {
```
olevett, line 247: Thanks for covering this off

olevett (approved):

PR #259, opened 2023-01-18
Feat: Update len lambda to update lev db

CarlyG55:


terraform/modules/len/lambdas/len.py
```diff
@@ -90,6 +76,83 @@ def post_event(auth_token: str):
...
+    connection = psycopg2.connect(user=lev_username, password=lev_password, database=lev_db_name, host=lev_host)
+    cursor = connection.cursor()
+    cursor.execute("select nextval('death_registration_ids')")
+    death_registration_id = cursor.fetchone()
+
+    cursor.execute("""
+        INSERT INTO death_registration_v1 (id, data, forenames, surname, date_of_birth, date_of_death)
+        VALUES(%(id)s, %(data)s, %(forenames)s, %(surname)s, %(date_of_birth)s, %(date_of_death)s)
+        """,
+                   generate_death_registration(death_registration_id))
```
CarlyG55, line 94: indentation seems a bit weird here but it's what intellij told me to do - I don't know python well enough to know if this is right

terraform/modules/lev_api/outputs.tf
```diff
@@ -1,3 +1,19 @@
...
+  value = aws_rds_cluster.lev_rds_postgres_cluster.master_password
+}
+
+output "lev_rds_db_name" {
+  value = aws_rds_cluster.lev_rds_postgres_cluster.database_name
+}
+
+output "lev_rds_db_host" {
+  value = aws_rds_cluster.lev_rds_postgres_cluster.endpoint
+}
```
CarlyG55, line 19: I'm still quite new to terraform, so this might be a very bad way to share credentials - very happy for input here 😅

lev-data/fix_lev_death_registration_ids.sql
```diff
@@ -0,0 +1,6 @@
+CREATE SEQUENCE death_registration_ids;
+
+UPDATE death_registration_v1
+SET id=nextval('death_registration_ids');
+UPDATE lev.public.death_registration_v1
+SET data = jsonb_set(data::jsonb, '{id}', to_jsonb(id));
```
ethanmills, line 6: I think I'd expect something like 
```sql
ALTER TABLE death_registration_v1 ALTER COLUMN id SET DEFAULT nextval('death_registration_ids');
```
here?

terraform/modules/len/lambdas/len.py
ethanmills: It's aligning on the open bracket. If you move the opening `"""` down a line it should look more normal

terraform/modules/len/lambdas/len.py
```diff
@@ -90,6 +76,83 @@ def post_event(auth_token: str):
...
+                   "district": "Test District",
+                   "administrativeArea": "Reading"
+               },
+               "informant": {
+                   "forenames": "Informy",
+                   "surname": "McInformface",
+                   "address": "666 Inform House, 6 Inform street, Informington, Informshire",
+                   "qualification": "Life-long nemesis, Present at death",
+                   "signature": "I. McInformface"
+               },
```
ethanmills, line 122: This gave me a chuckle, but I think it probably ought to be slightly more boring 😞 I know this was copied from the Home Office, but still...

terraform/modules/len/lambdas/psycopg2/__init__.py
```diff
@@ -0,0 +1,134 @@
+"""A Python driver for PostgreSQL
```
ethanmills, line 1: Unless I've missed it we don't have anything to zip this up do we?

terraform/modules/lev_api/outputs.tf
ethanmills: This mostly looks good to me. Might be worth adding a `sensitive = true` to the password

terraform/modules/len/lambdas/len.py
CarlyG55: Yeah I was wondering if it should be more 'normal' hahaha

terraform/modules/len/lambdas/len.py
CarlyG55: Nice, thanks!

terraform/modules/len/lambdas/psycopg2/__init__.py
CarlyG55: think it should be done correctly now

lev-data/fix_lev_death_registration_ids.sql
CarlyG55: Have added it in - although in the lambda we won't be using this default so the id's match up

PR #267, opened 2023-01-19
feat: Restrict access to the bastion host

olevett:


terraform/modules/data-share-service/ec2.tf
```diff
@@ -45,7 +45,7 @@ resource "aws_security_group" "rds_bastion_host_sg" {
     protocol    = "tcp"
     from_port   = 22
     to_port     = 22
-    cidr_blocks = ["0.0.0.0/0"]
+    cidr_blocks = var.externally_allowed_cidrs
```
olevett, line 48: Unfortunately, this doesn't mean we can remove the tfsec ingress rule, but it's definitely better

ethanmills: Awh why not?

olevett: I think because the IPs are still "public internet IPs" it's not a fan

PR #276, opened 2023-01-20
Feat: Remove inbound and outbound adaptors

CarlyG55:
In some places I feel I might have been a bit trigger happy with the delete key, in others I didn't delete something I wasn't sure about... so could probably do with a decent look over!

Wasn't sure whether there were things that needed removing from

- `ecs-security.tf`
- `update_security_groups.py`


README.md
```diff
@@ -1,9 +1,11 @@
 # GDX Death Notification POC
 
 ## API Contracts
+
```
CarlyG55, line 4: Annoyingly my formatter kept seeming to put these in even when I tried to remove them just before committing

src/main/kotlin/uk/gov/gdx/datashare/config/S3ClientsConfig.kt
```diff
@@ -5,7 +5,6 @@ import com.amazonaws.auth.BasicAWSCredentials
 import com.amazonaws.client.builder.AwsClientBuilder
```
olevett, line 5: I think we might be able to just delete tis whole thing, it looks like the AmazonS3 bean is only used in files that are deleted

olevett (approved):

olevett:
I think possibly worth @ethanmills  having a quick scan to check this won't cause problems with the consuming step function, but otherwise looks good to me

terraform/modules/data-share-service/s3.tf
```diff
@@ -66,22 +66,3 @@ resource "aws_s3_bucket_acl" "cloudfront_logs_bucket_grants" {
...
-  environment = var.environment
-  name        = "ingress-archive"
-}
-
-module "egress" {
-  source          = "../s3"
-  environment     = var.environment
-  name            = "egress"
-  expiration_days = 7
-}
```
ethanmills, line 87: I think the apply will fail here because these are not empty. We can either empty the buckets manually, or add a `force_destroy = true` here before removing them

ethanmills (approved):

src/main/kotlin/uk/gov/gdx/datashare/config/S3Config.kt
```diff
@@ -10,7 +10,4 @@ import org.springframework.validation.annotation.Validated
 data class S3Config(
   val localstackUrl: String = "",
```
Mjwillis, line 11: Do we need S3 config at all now?

Mjwillis (approved):
Looks good, I don't think we meed `S3Config` anymore either

Mjwillis (changes requested):
Think the S3 libraries can go from grade maybe?

src/main/kotlin/uk/gov/gdx/datashare/config/S3Config.kt
CarlyG55: Yeah wasn't sure about this - wasn't sure if it was needed for logs still

Mjwillis (approved):

olevett (approved):

PR #282, opened 2023-01-20
Fix: Fix len lambda bugs

CarlyG55:


lev-data/drop_not_nulls.sql
```diff
@@ -0,0 +1,8 @@
+/* These columns had not null constraints, but the corresponding columns in the
+   data json had null values. Removed constraints to keep new data consistent.
+ */
```
CarlyG55, line 3: Thought this might be worth a comment so we can come back to it after chatting to hmpo

olevett (approved):

ethanmills (approved):

PR #283, opened 2023-01-20
WIP: JSON:API (Reactor)

ethanmills:
TODO:

- [ ] Tests for pagination
- [ ] Tests for links
- [ ] POST /events format
- [ ] /events/status


Mjwillis (approved):
Feels like we should get this in today, even if we have to re-work after the reactive change

PR #286, opened 2023-01-23
feat: Remove reactive code

ethanmills:


src/main/kotlin/uk/gov/gdx/datashare/repository/ConsumerSubscriptionRepository.kt
```diff
@@ -1,21 +1,20 @@
...
 import java.util.*
 
 @Repository
-interface ConsumerSubscriptionRepository : CoroutineCrudRepository<ConsumerSubscription, UUID> {
+interface ConsumerSubscriptionRepository : CrudRepository<ConsumerSubscription, UUID> {
   @Query("SELECT * FROM consumer_subscription cs WHERE cs.oauth_client_id = :clientId")
-  fun findAllByClientId(clientId: String): Flow<ConsumerSubscription>
+  fun findAllByClientId(clientId: String): List<ConsumerSubscription>
 
   @Query(
```
Mjwillis, line 13: Could put JPA instead now

Mjwillis (approved):
Definitely worth discussing tomorrow, think my take is to switch. We have the option for JPA etc if we want then too and other libraries

build.gradle.kts
```diff
@@ -39,15 +35,14 @@ dependencies {
   implementation("org.hibernate:hibernate-validator:8.0.0.Final")
 
   implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core")
-  implementation("org.jetbrains.kotlinx:kotlinx-coroutines-reactor")
+  implementation("org.jetbrains.kotlinx:kotlinx-coroutines-reactive")
```
oskarwilliams378, line 38: Pls look at this

ethanmills: This is correct

PR #294, opened 2023-01-24
feat: Add audit table for GET /events

oskarwilliams378:


src/main/kotlin/uk/gov/gdx/datashare/repository/EventApiAudit.kt
```diff
@@ -0,0 +1,79 @@
...
+  val oauthClientId: String,
+  @Schema(description = "URL", required = true, example = "https://d33v84mi0vopmk.cloudfront.net/events")
+  val url: String,
+  @Schema(description = "METHOD", required = true, example = "GET")
+  val requestMethod: String,
+  @Schema(description = "Data of event(s)", required = true, example = "{\"data\":[]}")
+  val payload: Payload,
+
+  @Schema(description = "Time of call", required = true)
+  val whenCreated: LocalDateTime,
```
olevett, line 36: Is this actually time of call, or is this when we save it?

oskarwilliams378: Its when we save, will update the description

src/test/kotlin/uk/gov/gdx/datashare/controller/EventsControllerTest.kt
```diff
@@ -166,36 +168,6 @@ class EventsControllerTest {
     }
   }
 
-  @Test
```
olevett, line 169: Removing this test seems a little orthogonal to the actual change, is there a reason it doesn't work?

olevett (changes requested):
Couple of queries on some of the test related changes

src/test/kotlin/uk/gov/gdx/datashare/controller/EventsControllerTest.kt
oskarwilliams378: Its a duplicate test, it broke when I was doing this change which is why I noticed it

olevett (approved):

PR #296, opened 2023-01-24
feat: Revert S3 back to preserve

oskarwilliams378:


olevett:
Just to check, we did this to allow us to delete a couple of specific buckets, and they're now gone? Do we / should we parameterise this at the module layer?

PR #300, opened 2023-01-24
Document Death Event details in swagger

Mjwillis:
Map all V1 LEV fields for Death event
Hightlight fields available in swagger and in returned response


src/main/kotlin/uk/gov/gdx/datashare/service/EventDataService.kt
```diff
@@ -125,14 +126,21 @@ class EventDataService(
...
+    subscription: ConsumerSubscription,
+    includeData: Boolean = false,
+    callbackEvent: Boolean = false,
+  ): EventNotification {
     return EventNotification(
       eventId = event.id,
-      eventType = eventType,
+      eventType = subscription.eventType,
       sourceId = event.dataId,
+      dataIncluded = if (!callbackEvent) includeData else null,
```
Mjwillis, line 142: Indicates that data not / is included and which fields would be returned. These fields are only returned for `GET /events` polling not the `GET /events/{id}`

doc/architecture/decisions/0005-queue-api.md
```diff
@@ -107,7 +107,7 @@ Returns
       "eventType": "DEATH_NOTIFICATION",
       "sourceId": "{death certificate number}",
       "eventData": {
-        "firstName": "",
+        "firstNames": "",
```
olevett, line 110: I think HMPO use `forenames` are we trying to be consistent with that, or our own schema?

olevett: We'll also need to tell DWP about this as a breaking change

src/main/kotlin/uk/gov/gdx/datashare/service/DeathNotificationService.kt
```diff
@@ -98,16 +104,28 @@ class DeathNotificationService(
...
+  @Schema(description = "Date the death was registered", required = true, example = "2022-01-05", type = "date")
+  val registrationDate: LocalDate? = null,
+  @Schema(description = "Forenames of the deceased", required = true, example = "Bob Burt")
+  val firstNames: String? = null,
+  @Schema(description = "Surname of the deceased", required = true, example = "Smith")
   val lastName: String? = null,
-  @Schema(description = "Date of Birth", required = false, example = "2001-12-31T12:34:56")
-  val dateOfBirth: LocalDate? = null,
-  @Schema(description = "Date of Death", required = false, example = "2021-12-31T12:34:56")
+  @Schema(description = "Sex of the deceased", required = true, example = "Female", allowableValues = [ "Male", "Female", "Indeterminate"])
```
olevett, line 113: Not really in scope, but I'd be keen for this to be an enum

doc/architecture/decisions/0005-queue-api.md
Mjwillis: yes, I think we still should do it because its more accurate, what do you think?

src/main/kotlin/uk/gov/gdx/datashare/service/EventDataService.kt
```diff
@@ -161,11 +169,44 @@ data class EventNotification(
...
+  )
+  val dataIncluded: Boolean? = null,
+  @Schema(
+    description = "list of data fields that will be returned in this event",
+    required = false,
+    example = "firstNames,lastName,sex,dateOfDeath",
+  )
+  val enrichmentFields: String? = null,
+  @Schema(
+    description = "<h2>Event Data</h2>" +
```
olevett, line 184: I'd be keen for us to link out to the HMPO docs on this too to say "the authoritative source for this isn't us"

src/main/kotlin/uk/gov/gdx/datashare/controller/EventsController.kt
```diff
@@ -81,9 +81,28 @@ class EventsController(
...
+      "        \"dateOfDeath\": \"2023-01-02\",\n" +
+      "        \"dateOfBirth\": \"1972-02-20\",\n" +
+      "        \"birthPlace\": \"56 Test Address, B Town\",\n" +
+      "        \"deathPlace\": \"Hospital Ward 5, C Town\",\n" +
+      "        \"occupation\": \"Doctor\",\n" +
+      "        \"retired\": true,\n" +
+      "        \"address\": \"101 Address Street, A Town, Postcode\"\n" +
+      "}\n" +
+      "      </pre>" +
+      "      <p><b>Mandatory Fields</b>: registrationDate, firstNames, lastName, sex, dateOfDeath</p>",
```
olevett, line 105: With the mandatory, from what some of the HMPO examples have, I think it's "you'll always get this but it might be the empty string", not sure if we can clarify that?

src/main/kotlin/uk/gov/gdx/datashare/service/DeathNotificationService.kt
```diff
@@ -98,16 +104,28 @@ class DeathNotificationService(
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @Schema(description = "Death notification")
 data class DeathNotificationDetails(
```
olevett, line 106: There's probably a deeper future question for us to think about here which is does our service just pipe through the original entity, in which case calling it DeathNotificationDetails is right, or do we pipe through an event, with some citizen details from a variety of sources, in which case this structure doesn't feel quite right (maybe it's a `BaseEvent` with some standard citizen details and a couple of `DeathNotificationDetails` specific fields like `deathPlace`?)

olevett (approved):

src/main/kotlin/uk/gov/gdx/datashare/service/DeathNotificationService.kt
Mjwillis: Agree

src/main/kotlin/uk/gov/gdx/datashare/service/EventDataService.kt
Mjwillis: Good idea

src/main/kotlin/uk/gov/gdx/datashare/service/DeathNotificationService.kt
Mjwillis: Yes think we should discuss and agree way forward.  Maybe think about another event and what the `BaseEvent` looks like

PR #305, opened 2023-01-25
feat: Remove SNS configuration

olevett:
This was necessary, but since we're no longer using topics I figured it's easier than anything else to just remove


oskarwilliams378 (approved):
As long as this works, approved

Mjwillis (approved):
LGTM

PR #310, opened 2023-01-25
Update docs and GenderType enum

Mjwillis:


src/main/kotlin/uk/gov/gdx/datashare/service/DeathNotificationService.kt
```diff
@@ -145,3 +145,9 @@ data class DeathNotificationDetails(
   @Schema(description = "Whether the deceased was retired", required = false, example = "false")
   val retired: Boolean? = null,
 )
+
+enum class GenderType {
```
olevett, line 149:
```suggestion
enum class Gender {
```
not sure what "type" adds here, but happy to discuss

src/main/kotlin/uk/gov/gdx/datashare/service/DeathNotificationService.kt
```diff
@@ -145,3 +145,9 @@ data class DeathNotificationDetails(
...
+  Male,
```
olevett, line 150: There's a part of me that feels like enums should all be caps, but I assume this is a serialisation point thing?

Mjwillis: @olevett updated

src/main/kotlin/uk/gov/gdx/datashare/service/DeathNotificationService.kt
Mjwillis: @olevett updated

olevett (approved):

PR #311, opened 2023-01-25
feat: Remove data payload from egress

CarlyG55:


Mjwillis (approved):
LGTM

src/main/kotlin/uk/gov/gdx/datashare/service/DataProcessor.kt
```diff
@@ -23,37 +24,23 @@ class DataProcessor(
     val dataProcessorMessage: DataProcessorMessage = objectMapper.readValue(message, DataProcessorMessage::class.java)
     log.info("Received event [{}] from [{}]", dataProcessorMessage.eventTypeId, dataProcessorMessage.publisher)
 
-    // lookup provider
-    val details = getDataFromProvider(dataProcessorMessage)
+    val dataId = dataProcessorMessage.id ?: UUID.randomUUID().toString()
```
ethanmills, line 27: Probably a wider conversation, but is this valuable over just making this nullable?

src/main/kotlin/uk/gov/gdx/datashare/service/DataProcessor.kt
```diff
@@ -23,37 +24,23 @@ class DataProcessor(
...
 
-    when (dataProcessorMessage.eventTypeId) {
-      "DEATH_NOTIFICATION" -> deathNotificationService.saveDeathNotificationEvents(
-        details,
-        dataProcessorMessage,
-      )
-
-      "LIFE_EVENT" -> print("x == 2")
+    if (dataProcessorMessage.datasetId != "DEATH_LEV" && dataProcessorMessage.datasetId != "PASS_THROUGH") {
```
ethanmills, line 29: Can we have passthrough?

src/test/kotlin/uk/gov/gdx/datashare/service/DeathNotificationServiceTest.kt
```diff
@@ -2,78 +2,77 @@ package uk.gov.gdx.datashare.service
...
     )
-  }
 
-  @Test
-  fun `mapDeathNotification maps string to empty DeathNotificationDetails`() {
-    val input = "{}"
+    every { levApiService.findDeathById(dataId.toInt()) }.returns(listOf(deathRecord))
 
-    val deathNotificationDetails = underTest.mapDeathNotification(input)
+    val enrichedPayload = underTest.getEnrichedPayload(dataId, datasetId, enrichmentFields)
```
ethanmills, line 58: Super nitpicky, but if we assert this isn't null can we drop all the question marks?

ethanmills (approved):

ethanmills (approved):

src/main/kotlin/uk/gov/gdx/datashare/service/DataProcessor.kt
CarlyG55: It was in the codebase beforehand so I moved it over to here, and seems to be in our db, but I'm not sure exactly what it's used for at this point

src/main/kotlin/uk/gov/gdx/datashare/service/DataProcessor.kt
CarlyG55: Yeah I think there's a wider conversation around what situation we don't have an id in - because we can't fetch data from the lev without the id - and in fact maybe the random uuid just hides the fact we don't have one in the db?

PR #312, opened 2023-01-25
feat: Soft delete events

oskarwilliams378:


src/main/kotlin/uk/gov/gdx/datashare/repository/EventDataRepository.kt
```diff
@@ -55,15 +58,20 @@ interface EventDataRepository : CrudRepository<EventData, UUID> {
...
-  fun findAllByConsumerName(name: String): List<EventData>
+  @Override
+  override fun findById(id: UUID): Optional<EventData> = findByBaseIdAndDeletedAtIsNull(id)
+  fun findByBaseIdAndDeletedAtIsNull(id: UUID): Optional<EventData>
+
+  @Override
+  override fun findAll(): List<EventData> = findAllByDeletedAtIsNull()
+  fun findAllByDeletedAtIsNull(): List<EventData>
+
+  @Query("UPDATE event_data SET deleted_at=now() WHERE id = :id")
```
Mjwillis, line 74: Might be better to pass in the date rather and the DB now()

src/main/kotlin/uk/gov/gdx/datashare/repository/EventDataRepository.kt
```diff
@@ -8,12 +9,12 @@ import java.util.*
 
 @Repository
 interface EventDataRepository : CrudRepository<EventData, UUID> {
-
   @Query(
```
Mjwillis, line 12: Be nice to use JPA eventually here an we can filters tables

Mjwillis (approved):
Minor comments

PR #316, opened 2023-01-25
feat: update getEvents lambda to support new response format

ethanmills:


terraform/modules/consumer/lambdas/get_events.py
```diff
@@ -26,26 +26,48 @@
...
+        event_ids.extend(page_ids)
+
     record_metric(cloudwatch, cloudwatch_namespace, "GET_EVENTS.Duration", delta, unit="Seconds")
-    record_metric(cloudwatch, cloudwatch_namespace, "GET_EVENTS.Count", len(events))
+    record_metric(cloudwatch, cloudwatch_namespace, "GET_EVENTS.Count", total_events)
 
-    logger.info(f"Received {len(events)} events in {delta}s")
+    logger.info(f"Received {len(event_ids)} events in {delta}s")
+    if total_events != len(event_ids):
+        logger.error(f"Expected {total_events} events but received {len(event_ids)}")
```
ethanmills, line 57: This is possible because of the nature of limit-offset pagination, but it's worth keeping an eye on

olevett (approved):

PR #324, opened 2023-01-26
feat: Remove javax persistance

oskarwilliams378:


src/main/kotlin/uk/gov/gdx/datashare/repository/PublisherSubscription.kt
```diff
@@ -23,7 +21,6 @@ data class PublisherSubscription(
   val publisherId: UUID,
   @Schema(description = "Client ID", required = true, example = "a-client-id")
   val clientId: String,
-  @Enumerated(EnumType.STRING)
```
olevett, line 26: Do we still end up with strings in the database, given this?

oskarwilliams378: Yep

olevett (approved):

PR #338, opened 2023-01-27
docs: Add acquiring data docs

oskarwilliams378:


doc/how-to-onboard-a-client.md
```diff
@@ -81,7 +81,7 @@ Returns 200
 }
 ```
 
-### Tell the Consumer what to do
+## Tell the Consumer what to do
 
 The Consumer will now need to actually use this to access our API. For this they will need to be able to get themselves
 an access token, and then they will need to use that with the API.
```
ethanmills, line 87: Link to the new docs here?

ethanmills (approved):

doc/how-to-onboard-a-client.md
oskarwilliams378: Good plan, updating

Mjwillis (approved):
LGTM

PR #348, opened 2023-01-30
GPC-133: Add cognito client creation for admins

oskarwilliams378:


src/main/kotlin/uk/gov/gdx/datashare/controllers/AdminController.kt
```diff
@@ -31,4 +37,25 @@ class AdminController(
...
   fun getEvents() = eventDataRepository.findAll().toList()
+
+  @PostMapping("/cognitoClients")
+  @Operation(
+    summary = "Create cognito client",
+    description = "Need scope of events/admin",
+    responses = [
+      ApiResponse(
+        responseCode = "200",
+        description = "Event",
```
ethanmills, line 48: Is this correct?

src/main/kotlin/uk/gov/gdx/datashare/services/CognitoService.kt
```diff
@@ -0,0 +1,78 @@
...
+import uk.gov.gdx.datashare.models.CognitoClientRequest
+import uk.gov.gdx.datashare.models.CognitoClientResponse
+
+@Service
+class CognitoService(
+  @Value("\${cognito.user-pool-id}") val userPoolId: String,
+  @Value("\${cognito.enabled}") val enabled: Boolean,
+  @Value("\${cognito.acquirer-scope}") val acquirerScope: String,
+  @Value("\${cognito.supplier-scope}") val supplierScope: String,
+  @Value("\${cognito.admin-scope}") val adminScope: String,
```
ethanmills, line 20: Can we make these nullable and default to null so we don't need to set the in application-dev? Seems weird to need to set dummy values

src/main/kotlin/uk/gov/gdx/datashare/controllers/AdminController.kt
oskarwilliams378: Nope, thanks!

src/main/kotlin/uk/gov/gdx/datashare/services/CognitoService.kt
oskarwilliams378: Sure

PR #352, opened 2023-01-30
Create enrichment fields table

CarlyG55:


src/main/kotlin/uk/gov/gdx/datashare/services/ConsumersService.kt
```diff
@@ -14,45 +15,96 @@ import java.util.*
...
         consumerSubscriptionRepository.findByIdOrNull(subscriptionId)?.copy(
           consumerId = consumerId,
           oauthClientId = oauthClientId,
           eventType = eventType,
-          enrichmentFields = enrichmentFields,
           enrichmentFieldsIncludedInPoll = enrichmentFieldsIncludedInPoll,
         ) ?: throw ConsumerSubscriptionNotFoundException("Subscription $subscriptionId not found"),
       )
+
+      consumerSubscriptionEnrichmentFieldRepository.deleteAllByConsumerSubscriptionId(consumerSubscription.consumerSubscriptionId)
```
oskarwilliams378, line 103: Instead of deleting all of these, could we delete only the extra and update with the new ones, otherwise its a bit too much. We could also do all this update in a single query if you would rather write SQL

src/main/resources/db/devonly/R__dev_data.sql
```diff
@@ -49,18 +49,22 @@ INSERT INTO publisher_subscription
...
 INSERT INTO consumer_subscription
-(oauth_client_id, consumer_id, enrichment_fields, event_type, enrichment_fields_included_in_poll)
+(oauth_client_id, consumer_id, event_type, enrichment_fields_included_in_poll)
 VALUES ('hmrc-client', getIdFromConsumerName('Pub/Sub Consumer'),
-        'registrationDate,firstNames,lastName,maidenName,dateOfDeath,dateOfBirth,sex,address,birthplace,deathplace,occupation,retired',
         'DEATH_NOTIFICATION', true),
-       ('hmrc-client', getIdFromConsumerName('Pub/Sub Consumer'), '', 'LIFE_EVENT', true);
+       ('hmrc-client', getIdFromConsumerName('Pub/Sub Consumer'), 'LIFE_EVENT', true);
+
+INSERT INTO consumer_subscription_enrichment_field(consumer_subscription_id, enrichment_field)
```
oskarwilliams378, line 63: We'll need to delete this as well please :)

src/main/resources/db/migration/V1_25__consumer_subscription_enrichment_field.sql
```diff
@@ -0,0 +1,17 @@
...
+    enrichment_field         VARCHAR(200)     NOT NULL,
+
+    CONSTRAINT fk_consumer_subscription
+        FOREIGN KEY (consumer_subscription_id)
+            REFERENCES consumer_subscription (id)
+);
+
+INSERT INTO consumer_subscription_enrichment_field (consumer_subscription_id, enrichment_field)
+SELECT id, unnest(string_to_array(enrichment_fields::TEXT, ','))
+from consumer_subscription;
```
oskarwilliams378, line 14:
```suggestion
FROM consumer_subscription;
```

oskarwilliams378 (changes requested):

src/main/kotlin/uk/gov/gdx/datashare/services/ConsumersService.kt
CarlyG55: I chatted to olly about this, and he reckoned it would be better to delete and add them all, as the logic is a bit cleaner and it's still just two calls

src/main/resources/db/migration/V1_25__consumer_subscription_enrichment_field.sql
CarlyG55: whoops

src/main/kotlin/uk/gov/gdx/datashare/services/ConsumersService.kt
oskarwilliams378: Sure, SGTM :)

src/main/resources/db/devonly/R__dev_data.sql
```diff
@@ -44,23 +44,33 @@ DELETE
 FROM consumer_subscription
 WHERE oauth_client_id IN ('dwp-event-receiver', 'hmrc-client', 'internal-outbound');
 
+DELETE
+FROM consumer_subscription_enrichment_field
+WHERE enrichment_field IN
```
oskarwilliams378, line 49: Could we only add and delete for the consumer subscriptions added in this. Also could we move this above the delete from consumer subscriptions

oskarwilliams378 (changes requested):

oskarwilliams378 (approved):

PR #356, opened 2023-01-30
GPC-135 acquirer creation

oskarwilliams378:


src/main/kotlin/uk/gov/gdx/datashare/controllers/AdminController.kt
```diff
@@ -59,4 +66,39 @@ class AdminController(
...
+      consumer.id,
+      ConsumerSubRequest(
+        oauthClientId = it.clientId,
+        eventType = createAcquirerRequest.eventType,
+        enrichmentFields = createAcquirerRequest.enrichmentFields,
+        enrichmentFieldsIncludedInPoll = createAcquirerRequest.enrichmentFieldsIncludedInPoll,
+      ),
+    )
+
+    CognitoClientResponse(it.clientName, it.clientId, it.clientSecret)
```
ethanmills, line 102: I think  this should be in a service somewhere, and marked with `@Transactional` so we don't end up with half complete data

oskarwilliams378: Done

src/main/kotlin/uk/gov/gdx/datashare/services/AdminService.kt
```diff
@@ -0,0 +1,33 @@
...
+  private val consumersService: ConsumersService,
+) {
+  fun createCognitoClient(cognitoClientRequest: CognitoClientRequest) =
+    cognitoService.createUserPoolClient(cognitoClientRequest)
+
+  @Transactional
+  fun createAcquirer(createAcquirerRequest: CreateAcquirerRequest) = cognitoService.createUserPoolClient(
+    CognitoClientRequest(createAcquirerRequest.clientName, listOf(CognitoClientType.ACQUIRER)),
+  )?.let {
+    val consumer = consumersService.addConsumer(ConsumerRequest(createAcquirerRequest.clientName))
```
ethanmills, line 20: Do we have a uniqueness constraint on the client name? If so, would be worth moving above the cognito creation so we fail fast.

ethanmills (approved):

PR #362, opened 2023-01-31
GPC-NONE: Rename publisher to supplier

oskarwilliams378:


src/main/kotlin/uk/gov/gdx/datashare/services/DataReceiverService.kt
```diff
@@ -59,7 +59,7 @@ class DataReceiverService(
     log.debug(
       "Notifying Data Processor of event type {} from {}",
       dataProcessorMessage.eventType,
-      publisher.name,
+      supplier.name,
     )
 
     meterRegistry.counter("EVENT_ACTION.EventPublished", "eventType", eventPayload.eventType.name).increment()
```
ethanmills, line 65: I assume we want to keep this over e.g. EventSupplied?

ethanmills (approved):

src/main/kotlin/uk/gov/gdx/datashare/services/DataReceiverService.kt
oskarwilliams378: I thought so, but we can change later if we want

PR #365, opened 2023-01-31
GPC-78: add api versioning

CarlyG55:


.github/workflows/build-and-push.yml
```diff
@@ -71,7 +86,7 @@ jobs:
     if: inputs.push
     strategy:
       matrix:
-        env: [dev, demo]
+        env: [ dev, demo ]
```
olevett, line 89: Your yaml linter is very agressive!

olevett:
This seems fine, to check this is going to give us incrementing builds, one per commit since the 0.0.1 tag?

olevett (approved):

.github/workflows/build-and-push.yml
CarlyG55: I don't even know where it's come from 😅

CarlyG55:
> This seems fine, to check this is going to give us incrementing builds, one per commit since the 0.0.1 tag? I think I'd be keen to put _some_ information in the tag, like `version-0.0.1`, and tell GitVersion what the prefix is, but I'm not too worried by this.

From what I can tell, yes - I'll give this a go

PR #366, opened 2023-01-31
GPC-144: Add client name validation

oskarwilliams378:


src/main/kotlin/uk/gov/gdx/datashare/models/CreateAcquirerRequest.kt
```diff
@@ -1,10 +1,23 @@
...
-  @Schema(description = "Name of client, must be lower case", required = true, example = "HMPO")
+class CreateAcquirerRequest(
+  @Schema(
+    description = "Name of client, may only contain lowercase letters, numbers, and the following special characters: + = , . @ -",
+    required = true,
+    example = "hmpo",
+    maxLength = 128,
+    pattern = "^[a-z0-9+=,.@-]*\$",
+  )
+  @get:Length(min = 3, max = 128)
```
olevett, line 16: Is it worth writing _somewhere_ why these constraints exist? E.g. if we moved off of Cognito, would these change?

src/main/kotlin/uk/gov/gdx/datashare/models/CreateSupplierRequest.kt
```diff
@@ -1,10 +1,23 @@
...
+    example = "hmpo",
+    maxLength = 128,
+    pattern = "^[a-z0-9+=,.@-]*\$",
+  )
+  @get:Length(min = 3, max = 128)
+  @get:Pattern(
+    regexp = "^[a-z0-9+=,.@-]*\$",
+    message = "Name may only contain lowercase letters, numbers, and the following special characters: + = , . @ -",
+  )
   val clientName: String,
```
olevett, line 21: That we've written the same set of constraints twice, is I think the strongest reason I could come up with for pulling this into a type. I'm not too worried here, but I think it's worth keeping in mind.

olevett (approved):

PR #374, opened 2023-01-31
GPC-137 On board new supplier (MOJ)

Mjwillis:


oauth2-data/config-local.json
```diff
@@ -42,6 +42,15 @@
             "client_id": "hmrc-client"
           }
         },
+        {
```
Mjwillis, line 45: This isn't needed in the real application as its a queue

src/main/kotlin/uk/gov/gdx/datashare/services/PrisonerEventMessageProcessor.kt
```diff
@@ -0,0 +1,86 @@
...
+      "prison-offender-events.prisoner.received" -> processHMPPSEvent(hmppsDomainEvent)
+      else -> {
+        log.debug("Ignoring message with type $eventType")
+      }
+    }
+  }
+
+  fun processHMPPSEvent(hmppsDomainEvent: HMPPSDomainEvent) {
+    // find a supplier with a defined name
+    supplierSubscriptionRepository.findAllByEventType(EventType.ENTERED_PRISON)
```
Mjwillis, line 45: Just find any supplier who supports this event type

src/main/kotlin/uk/gov/gdx/datashare/config/WebClientConfiguration.kt
```diff
@@ -37,4 +42,47 @@ class WebClientConfiguration(
...
+
+    val connectionProvider = ConnectionProvider.builder("prisonerSearch")
+      .maxConnections(100)
+      .maxIdleTime(Duration.ofSeconds(20))
+      .maxLifeTime(Duration.ofSeconds(60))
+      .pendingAcquireTimeout(Duration.ofSeconds(60))
+      .evictInBackground(Duration.ofSeconds(120))
+      .build()
+
+    val httpClient = HttpClient.create(connectionProvider).responseTimeout(Duration.ofMinutes(10))
```
oskarwilliams378, line 60: can we extract the http client creation to a seperate function as it is the same for both webclients

src/main/kotlin/uk/gov/gdx/datashare/controllers/EventsController.kt
```diff
@@ -156,7 +156,6 @@ class EventsController(
     @Schema(
       description = "Event Types, if none supplied it will be the allowed types for this client",
       required = false,
-      allowableValues = ["DEATH_NOTIFICATION", "LIFE_EVENT"],
```
oskarwilliams378, line 159: Why remove this all over?

src/main/kotlin/uk/gov/gdx/datashare/enums/EventType.kt
```diff
@@ -3,4 +3,5 @@ package uk.gov.gdx.datashare.enums
 enum class EventType {
   DEATH_NOTIFICATION,
   LIFE_EVENT,
+  ENTERED_PRISON,
```
oskarwilliams378, line 6: General question, will we be only receiving entered prison events?

src/main/kotlin/uk/gov/gdx/datashare/services/EventDataService.kt
```diff
@@ -119,20 +120,25 @@ class EventDataService(
...
-        )
-      } else {
-        null
-      },
+      enrichmentFields = enrichmentFields,
+      eventData = if (includeData) callbackAndEnrichData(subscription, event, enrichmentFields) else null,
     )
   }
+
+  @Suppress("IMPLICIT_CAST_TO_ANY")
```
oskarwilliams378, line 134: any reason for this vs

```
: Any?
```

on the function return?

src/main/kotlin/uk/gov/gdx/datashare/services/PrisonerApiService.kt
```diff
@@ -0,0 +1,41 @@
...
+import uk.gov.gdx.datashare.config.NoDataFoundException
+import uk.gov.gdx.datashare.models.PrisonerRecord
+
+@Service
+@XRayEnabled
+class PrisonerApiService(
+  private val prisonerSearchApiWebClient: WebClient,
+  meterRegistry: MeterRegistry,
+) {
+  private val callsToPrisonerSearchCounter: Counter = meterRegistry.counter("API_CALLS.CallsToPrisonerSearch")
```
oskarwilliams378, line 19: Nothing to do with this, but I think we should change our naming scheme for metrics

src/main/kotlin/uk/gov/gdx/datashare/services/PrisonerEventMessageProcessor.kt
```diff
@@ -0,0 +1,86 @@
...
+    additionalInformation,
+    occurredAt.toOffsetDateFormat(),
+  )
+}
+
+data class AdditionalInformation(
+  val nomsNumber: String,
+)
+fun Instant.toOffsetDateFormat(): String =
+  atZone(ZoneId.of("Europe/London")).toOffsetDateTime().format(DateTimeFormatter.ISO_OFFSET_DATE_TIME)
```
oskarwilliams378, line 86: could we add this to the datetimehelper or something, just so it can be shared :)

src/main/resources/application-dev.yml
```diff
@@ -31,14 +31,24 @@ spring:
...
+      prisoner-search: http://localhost:8082
+      hmpps-auth: http://localhost:9090/issuer1
 
 xray-tracing:
   enabled: false
 
 cognito:
   enabled: false
+
+
```
oskarwilliams378, line 54: double new line :D

src/main/resources/db/devonly/R__dev_data.sql
```diff
@@ -75,5 +75,23 @@ FROM acquirer_subscription
 WHERE event_type = 'DEATH_NOTIFICATION'
   AND oauth_client_id IN ('dwp-event-receiver', 'hmrc-client');
 
+
+INSERT INTO supplier (name) VALUES ('HMPPS');
```
oskarwilliams378, line 79: Could we also delete the things we add :)

src/main/kotlin/uk/gov/gdx/datashare/services/PrisonerEventMessageProcessor.kt
oskarwilliams378: is this for enrichment? Or just a check that the queue is only running if we have set up HMPPS as a supplier?

Mjwillis: Just a check that we have a supplier that handles that event type

src/main/kotlin/uk/gov/gdx/datashare/config/WebClientConfiguration.kt
Mjwillis: yes!

src/main/kotlin/uk/gov/gdx/datashare/controllers/EventsController.kt
Mjwillis: Because its now an enum and you get it for free!

src/main/kotlin/uk/gov/gdx/datashare/enums/EventType.kt
Mjwillis: We can get LEAVING PRISON or lots of other if we wanted

src/main/kotlin/uk/gov/gdx/datashare/services/EventDataService.kt
Mjwillis: No! - Will change

src/main/kotlin/uk/gov/gdx/datashare/controllers/EventsController.kt
oskarwilliams378: Huh I didn't see it in swagger for free...

src/main/resources/db/devonly/R__dev_data.sql
Mjwillis: Sorry not sure what you mean!

ethanmills: This is a repeatable migration, so it should be idempotent. Something like `DELETE FROM supplier WHERE NAME = 'HMPSS';` before this.

src/main/kotlin/uk/gov/gdx/datashare/config/WebClientConfiguration.kt
```diff
@@ -22,19 +27,57 @@ class WebClientConfiguration(
...
+    val connectionProvider = ConnectionProvider.builder(connectionName)
       .maxConnections(100)
       .maxIdleTime(Duration.ofSeconds(20))
       .maxLifeTime(Duration.ofSeconds(60))
       .pendingAcquireTimeout(Duration.ofSeconds(60))
       .evictInBackground(Duration.ofSeconds(120))
       .build()
-    val httpClient = HttpClient.create(connectionProvider).responseTimeout(Duration.ofSeconds(10))
+
+    return HttpClient.create(connectionProvider).responseTimeout(Duration.ofMinutes(10))
```
ethanmills, line 61: I think a 10 minute timeout here is pretty extreme? It's also incompatible with the other configured options I think

src/main/kotlin/uk/gov/gdx/datashare/services/EventDataService.kt
```diff
@@ -119,20 +120,24 @@ class EventDataService(
...
+      enrichmentFields = enrichmentFields,
+      eventData = if (includeData) callbackAndEnrichData(subscription, event, enrichmentFields) else null,
     )
   }
+
+  private fun callbackAndEnrichData(
+    subscription: AcquirerSubscription,
+    event: EventData,
+    enrichmentFields: List<String>,
+  ): Any? = when (subscription.eventType) {
```
ethanmills, line 138: Perhaps not now, but we should find a way to type this better

src/main/kotlin/uk/gov/gdx/datashare/services/PrisonerEventMessageProcessor.kt
```diff
@@ -0,0 +1,86 @@
...
+) {
+  companion object {
+    val log: Logger = LoggerFactory.getLogger(this::class.java)
+  }
+
+  @JmsListener(destination = "prisoner-event", containerFactory = "awsQueueContainerFactoryProxy")
+  fun onPrisonerEventMessage(prisonerEventMessage: String) {
+    val (message, messageAttributes) = objectMapper.readValue(prisonerEventMessage, TopicMessage::class.java)
+    val eventType = messageAttributes.eventType.Value
+    log.info("Received message $message, type $eventType")
```
ethanmills, line 32: logging PII?

src/main/kotlin/uk/gov/gdx/datashare/services/PrisonerEventMessageProcessor.kt
ethanmills: Perhaps a premature optimisation, but we could use `findFirstByEventType` here

src/main/kotlin/uk/gov/gdx/datashare/services/PrisonerEventMessageProcessor.kt
```diff
@@ -0,0 +1,86 @@
...
+    // find a supplier with a defined name
+    supplierSubscriptionRepository.findAllByEventType(EventType.ENTERED_PRISON)
+      .firstOrNull()?.let {
+        dataReceiverService.sendToDataProcessor(
+          EventToPublish(
+            EventType.ENTERED_PRISON,
+            LocalDateTime.parse(hmppsDomainEvent.occurredAt, DateTimeFormatter.ISO_OFFSET_DATE_TIME),
+            hmppsDomainEvent.additionalInformation.nomsNumber,
+          ),
+          it.clientId,
```
ethanmills, line 53: Not sure if I like dynamically setting the client ID like this. We know this is always coming from HMPPS, right? I wonder if there are any downsides to doing it this way

src/main/kotlin/uk/gov/gdx/datashare/services/PrisonerLookupService.kt
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  fun getEnrichedPayload(
+    prisonerNumber: String,
+    enrichmentFields: List<String>,
+  ): PrisonerDetails? {
+    val allEnrichedData = prisonerApiService.findPrisonerById(prisonerNumber)
+      ?.let {
+        PrisonerDetails(
+          prisonerNumber = prisonerNumber,
```
ethanmills, line 28: I assume the prisonerNumber is guaranteed never to change and "redirect"?

src/main/kotlin/uk/gov/gdx/datashare/config/WebClientConfiguration.kt
oskarwilliams378: Updated to 10s

src/main/kotlin/uk/gov/gdx/datashare/services/EventDataService.kt
oskarwilliams378: Agree that we should later on

src/main/kotlin/uk/gov/gdx/datashare/services/PrisonerEventMessageProcessor.kt
oskarwilliams378: Have changed to log noms as we are currently saving that as non PII

src/main/kotlin/uk/gov/gdx/datashare/services/PrisonerEventMessageProcessor.kt
oskarwilliams378: I think this is a question we need to solve about using queues as inputs rather than oauth'd users

src/main/kotlin/uk/gov/gdx/datashare/services/PrisonerLookupService.kt
oskarwilliams378: I assume so

Mjwillis: correct

src/main/kotlin/uk/gov/gdx/datashare/services/PrisonerEventMessageProcessor.kt
Mjwillis: agreed

Mjwillis: Was just doing this at this point to hook into existing process for adding an event

src/main/resources/db/devonly/R__dev_data.sql
Mjwillis: 👍

Mjwillis:
@olevett / @ethanmills @oskarwilliams378 - The enum applied to enrichment fields work that @CarlyG55 picked up completely clobbers my PR - we've hard coded the enum into the aquirer subscription for which a prisoner entering prison has completely different fields!  Will have to pick this up next week, but suspect a specific enum on the database table doesn't work

src/main/resources/db/devonly/R__dev_data.sql
Mjwillis: Done

oskarwilliams378:
@Mjwillis FYI we have had a chat about this today, and the change to enrichment field as is is what we plan to continue with for the moment

oskarwilliams378 (approved):

PR #383, opened 2023-02-01
GPC-104: Add cddo openapi linting

oskarwilliams378:


.github/workflows/openapi-lint.yml
```diff
@@ -0,0 +1,39 @@
...
+        with:
+          node-version: 18
+
+      - name: Spin up localstack and db
+        uses: yu-ichiro/spin-up-docker-compose-action@v1
+        with:
+          file: docker-compose-local.yml
+
+      - name: Install packages
+        run: npm ci
```
olevett, line 33: Does the generate step need the npm dependency, if not, I'd be tempted to put it down a bit closer to the test step

package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "devDependencies": {
+    "@stoplight/spectral-cli": "^6.2.0",
```
olevett, line 3: I think, given we've got dependabot etc, I'd make this a fixed dependency and lose the `^`, not too strongly opinionated here though

.github/workflows/openapi-lint.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - name: Generate API docs
+        run: ./gradlew -Pdocs generateOpenApiDocs
+
+      - name: Lint API docs
+        run: npm run openapi-lint
```
olevett, line 39: Do we get any concrete output of this, or is it just a pass/fail?

package.json
oskarwilliams378: Sure

.github/workflows/openapi-lint.yml
oskarwilliams378: Will do

package.json
ethanmills: do we need to configure dependabot for package.json?

oskarwilliams378: looks like we may not need package at all I hope

.github/workflows/openapi-lint.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Spin up localstack and db
+        uses: yu-ichiro/spin-up-docker-compose-action@v1
+        with:
+          file: docker-compose-local.yml
+
+      - name: Generate API docs
+        run: ./gradlew -Pdocs generateOpenApiDocs
+
+      - name: Lint API docs
+        uses: stoplightio/spectral-action@latest
```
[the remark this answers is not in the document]
oskarwilliams378: I'm getting it working then will lock it down

olevett (approved):

PR #389, opened 2023-02-02
GPC-145: Change event response to always include allowed fields

ethanmills:


src/main/kotlin/uk/gov/gdx/datashare/models/DeathNotificationDetails.kt
```diff
@@ -52,4 +57,72 @@ data class DeathNotificationDetails(
...
+        )
+      }
+    }
+  }
+}
+
+class DeathNotificationDetailsSerializer :
+  StdSerializer<DeathNotificationDetails>(DeathNotificationDetails::class.java) {
+  override fun serialize(value: DeathNotificationDetails?, gen: JsonGenerator?, provider: SerializerProvider?) {
+    gen?.writeStartObject()
```
olevett, line 87: not having a generator here feels like a problem, can we check it for null once and then just use it throughout?

src/main/kotlin/uk/gov/gdx/datashare/models/DeathNotificationDetails.kt
```diff
@@ -52,4 +57,72 @@ data class DeathNotificationDetails(
...
+    value?.run {
+      if (enrichmentFields.contains("registrationDate")) {
```
olevett, line 89: I can't help thinking there's got to be a way to make this less branchy logic, but I'm not sure at all what that is :)

src/main/kotlin/uk/gov/gdx/datashare/models/DeathNotificationDetails.kt
```diff
@@ -52,4 +57,72 @@ data class DeathNotificationDetails(
...
+        gen?.writeNullableStringField("firstNames", firstNames)
+      }
+      if (enrichmentFields.contains("lastName")) {
+        gen?.writeNullableStringField("lastName", lastName)
+      }
+      if (enrichmentFields.contains("sex")) {
+        gen?.writeNullableStringField("sex", sex?.jsonName)
+      }
+      if (enrichmentFields.contains("dateOfDeath")) {
+        gen?.writeNullableStringField("dateOfDeath", dateOfDeath?.toString())
```
olevett, line 102: Do these serialize as we'd expect or should we format the date ourselves?

olevett: For example would it be better to do this date->string/json mapping within our writeNullableStringField bit instead?

olevett (approved):

src/main/kotlin/uk/gov/gdx/datashare/models/DeathNotificationDetails.kt
ethanmills: Yeah this comes out fine. It gets messy when you start adding times, but date works

PR #394, opened 2023-02-02
GPC-129: Set lev api credentials in aws ssm

CarlyG55:


ethanmills:
Do we need some IAM policy to allow the ECS task execution role to access SSM params?

build.gradle.kts
```diff
@@ -97,6 +98,7 @@ dependencies {
   testImplementation("org.mockito.kotlin:mockito-kotlin:4.1.0")
   testImplementation("org.mockito:mockito-inline:5.1.1")
   testImplementation("com.approvaltests:approvaltests:18.5.0")
+  testImplementation("com.ninja-squad:springmockk:3.0.0")
```
ethanmills, line 101: Shouldn't need this any more

src/test/kotlin/uk/gov/gdx/datashare/TestApplication.kt
```diff
@@ -0,0 +1,30 @@
...
+import org.springframework.context.annotation.Primary
+import software.amazon.awssdk.services.ssm.SsmClient
+import software.amazon.awssdk.services.ssm.model.GetParameterResponse
+import software.amazon.awssdk.services.ssm.model.Parameter
+
+@SpringBootApplication
+class TestApplication {
+  @Bean
+  @Primary
+  fun mockSsmClient(): SsmClient {
```
ethanmills, line 17: This has to be `ssmClient`, or you need to use `@Bean(name="ssmClient")`

ethanmills (changes requested):

ethanmills (approved):

PR #404, opened 2023-02-03
GPC-137 Added environment variables

Mjwillis:


terraform/modules/data-share-service/variables.tf
```diff
@@ -4,7 +4,16 @@ variable "ecr_url" {}
...
   type = list(string)
 }
+
+variable "prisoner-event-enabled" {}
+variable "prisoner-search-url" {}
+variable "hmpps-auth_url" {}
+variable "prisoner-search-client-id" {}
+variable "prisoner-search-client-secret" {}
+variable "prisoner-event-queue-name" {}
+variable "prisoner-event-dlq-name" {}
```
oskarwilliams378, line 19: Could we change these from kebab to snake case please :)

oskarwilliams378 (changes requested):

terraform/modules/data-share-service/variables.tf
Mjwillis: done!

oskarwilliams378 (approved):

PR #405, opened 2023-02-03
GPC-95: Restrict ECS egress to HTTPS

oskarwilliams378:


terraform/modules/data-share-service/ecs-security.tf
```diff
@@ -13,9 +13,9 @@ resource "aws_security_group" "ecs_tasks" {
   }
 
   egress {
-    protocol    = "-1"
-    from_port   = 0
-    to_port     = 0
+    protocol    = "tcp"
+    from_port   = 443
```
ethanmills, line 17: Are you sure this is correct? I think outbound connections come from a random port

terraform/modules/data-share-service/ecs-security.tf
```diff
@@ -13,9 +13,9 @@ resource "aws_security_group" "ecs_tasks" {
   }
 
   egress {
```
ethanmills, line 15: Is this just to outside the VPC? For example we need 5432 for postgres - will that work?

terraform/modules/data-share-service/ecs-security.tf
oskarwilliams378: this defines the ranfe

ethanmills (approved):

PR #421, opened 2023-02-06
GPC-156 Add SSM config for Queue and other secrets

Mjwillis:


terraform/modules/data-share-service/ssm.tf
```diff
@@ -9,3 +9,27 @@ resource "aws_ssm_parameter" "lev_api_client_user" {
...
+resource "aws_ssm_parameter" "prisoner_search_api_client_secret" {
+  name  = "${var.environment}_prisoner_search_api_client_secret"
+  type  = "SecureString"
+  value = "dummyvalue"
+}
+
+resource "aws_ssm_parameter" "prisoner_event_queue_name" {
+  name  = "${var.environment}_prisoner_event_queue_name"
+  type  = "String"
+  value = "dummyvalue"
```
oskarwilliams378, line 28: do you want to preseed this with the correct value?

Mjwillis: Yes probably should - what happens if we over write manually does terraform ever update again?

src/main/kotlin/uk/gov/gdx/datashare/config/WebClientConfiguration.kt
```diff
@@ -21,7 +21,8 @@ import java.time.Duration
 class WebClientConfiguration(
   @Value("\${api.base.url.lev}") private val levApiRootUri: String,
   @Value("\${api.base.url.prisoner-search:-}") private val prisonerSearchApiUri: String,
-  @Value("\${environment:-}") private val environment: String,
+  @Value("\${api.base.lev.api.client.name:-}") private val levApiClientName: String,
```
Mjwillis, line 24: Remove need to inject env and all that SSM client stuff

src/main/kotlin/uk/gov/gdx/datashare/queue/AwsQueue.kt
```diff
@@ -9,7 +9,16 @@ class AwsQueue(
...
   val sqsDlqClient: SqsClient? = null,
   val dlqName: String? = null,
+  private val awsAccountId: String? = null,
 ) {
-  val queueUrl: String by lazy { sqsClient.getQueueUrl(GetQueueUrlRequest.builder().queueName(queueName).build()).queueUrl() }
-  val dlqUrl by lazy { sqsDlqClient?.getQueueUrl(GetQueueUrlRequest.builder().queueName(dlqName).build())?.queueUrl() }
+  val queueUrl: String by lazy { sqsClient.getQueueUrl(buildQueueUrlRequest(queueName, awsAccountId)).queueUrl() }
+  val dlqUrl by lazy { sqsDlqClient?.getQueueUrl(buildQueueUrlRequest(dlqName, awsAccountId))?.queueUrl() }
+
+  private fun buildQueueUrlRequest(name: String?, accountId: String?): GetQueueUrlRequest {
```
Mjwillis, line 17: This is needed when queue name is a different AWS account

src/main/kotlin/uk/gov/gdx/datashare/queue/AwsQueueDestinationResolver.kt
```diff
@@ -10,4 +12,13 @@ class AwsQueueDestinationResolver(private val sqsProperties: SqsProperties) : Dy
     val destination = sqsProperties.enabledQueues[destinationName]?.queueName ?: destinationName
     return super.resolveDestinationName(session, destination, pubSubDomain)
   }
+
+  override fun resolveQueue(session: Session, queueName: String): Queue {
```
Mjwillis, line 16: override this method as does not handle separate account

Mjwillis:
@oskarwilliams378 I've removed all the SSM stuff as its over complex and not needed at this point

PR #423, opened 2023-02-06
GPC-93: Add e2e tests

oskarwilliams378:


src/main/kotlin/uk/gov/gdx/datashare/models/TestEvent.kt
```diff
@@ -0,0 +1,11 @@
+package uk.gov.gdx.datashare.models
+
+import com.fasterxml.jackson.annotation.JsonInclude
+import io.swagger.v3.oas.annotations.media.Schema
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@Schema(description = "Test event")
```
ethanmills, line 7: If we don't have a schema here does it not appear in the swagger? I'm not sure we need to publish this as part of our public API (and can therefore change it without it being a breaking change)

terraform/modules/data-share-service/cloudfront.tf
```diff
@@ -1,3 +1,7 @@
+locals {
+  gdx_url = "https://${aws_cloudfront_distribution.gdx_data_share_poc.domain_name}"
```
ethanmills, line 2: `gdx_api_base_url`?

terraform/modules/deploy_hook/lambdas/hook.py
```diff
@@ -0,0 +1,91 @@
...
+
+    put_lifecycle_event_hook(event, 'InProgress')
+
+    auth_token = get_auth_token(auth_url, client_id, client_secret)
+
+    post_event(auth_token)
+    events = get_events(auth_token)
+    get_event(auth_token, events[0]["id"])
+
+    put_lifecycle_event_hook(event, 'Succeeded')
```
ethanmills, line 38: Do we need to do error handling and explicitly put a failed event?

terraform/modules/deploy_hook/lambdas/hook.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def get_events(auth_token: str):
+    logger.info(f"## Getting test events")
+    events = get_data(auth_token, events_url)["data"]
+    logger.info(f"## Successfully retrieved {len(events)} test event(s)")
+    return events
+
+
+def get_event(auth_token: str, event_id: str):
```
ethanmills, line 78: Do we need to delete this event once we're done?

ethanmills (approved):

src/main/kotlin/uk/gov/gdx/datashare/models/TestEvent.kt
oskarwilliams378: Happy to remove

terraform/modules/deploy_hook/lambdas/hook.py
oskarwilliams378: Yes I should do that

terraform/modules/deploy_hook/lambdas/hook.py
oskarwilliams378: And this

PR #425, opened 2023-02-06
GPC-161: Add github oidc

oskarwilliams378:


terraform/modules/data-share-service/github_iam.tf
```diff
@@ -0,0 +1,81 @@
...
+  }
+}
+
+resource "aws_iam_role" "github_oidc_environment" {
+  name               = "${var.environment}-github-oidc-deploy"
+  assume_role_policy = data.aws_iam_policy_document.github_oidc_environment_assume.json
+}
+
+data "aws_iam_policy" "github_oidc_environment" {
+  name = "Admin"
```
ethanmills, line 32: I know this isn't a change, but I find this very scary

terraform/modules/data-share-service/github_iam.tf
```diff
@@ -0,0 +1,81 @@
...
+resource "aws_iam_role" "github_oidc_pull_request" {
+  count = var.environment == "dev" ? 1 : 0
+
+  name               = "github-oidc-pull-request"
+  assume_role_policy = data.aws_iam_policy_document.github_oidc_pull_request_assume.json
+}
+
+data "aws_iam_policy" "github_oidc_pull_request" {
+  name = "ReadOnlyAccess"
+}
```
ethanmills, line 74: Oh this is nice - so we get read only access when running PR changes, then full write when deploying? Nice!

.github/workflows/build-and-push.yml
```diff
@@ -53,8 +52,8 @@ jobs:
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@v1
         with:
-          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          role-to-assume: arn:aws:iam::776473272850:role/dev-github-oidc-deploy
```
ethanmills, line 55: We could probably lock this down further at some point to just a role that has e.g. ECR full access. Not now though 🙂

ethanmills (approved):
Presumably this may need two attempts at deploying?

terraform/modules/data-share-service/github_iam.tf
oskarwilliams378: I am happy for us to lock this down if you know what terraform needs

terraform/modules/data-share-service/github_iam.tf
oskarwilliams378: Yep!

PR #434, opened 2023-02-07
GPC-167 Add swagger docs for new Event type for Receiving a prisoner in Prison

Mjwillis:


src/main/kotlin/uk/gov/gdx/datashare/controllers/EventsController.kt
```diff
@@ -77,7 +77,20 @@ class EventsController(
       "}\n" +
       "      </pre>" +
       "      <p><b>Mandatory Fields</b>: registrationDate, firstNames, lastName, sex, dateOfDeath</p>" +
-      "      <p><b>Gender Types:</b>: Male, Female, Indeterminate</p>",
+      "      <p><b>Gender Types:</b>: Male, Female, Indeterminate</p>" +
```
olevett, line 80:
```suggestion
      "      <p><b>Sex:</b>: Male, Female, Indeterminate</p>" +
```

I think since HMPO return `sex` as the field we should be conscious of the specific terminology

olevett (approved):

src/main/kotlin/uk/gov/gdx/datashare/controllers/EventsController.kt
olevett: Oh, or if HMPO return Gender then I think we should be returning that

PR #437, opened 2023-02-07
GPC-168 Missing policy to access SSM secrets

Mjwillis:


terraform/modules/data-share-service/iam.tf
```diff
@@ -173,7 +173,12 @@ data "aws_iam_policy_document" "ecs_task_ssm_access" {
     actions = ["ssm:GetParameter"]
     resources = [
       aws_ssm_parameter.lev_api_client_name.arn,
-      aws_ssm_parameter.lev_api_client_user.arn
+      aws_ssm_parameter.lev_api_client_user.arn,
+      aws_ssm_parameter.prisoner_search_api_client_id.arn,
+      aws_ssm_parameter.prisoner_search_api_client_secret.arn,
+      aws_ssm_parameter.prisoner_event_queue_name.arn,
```
ethanmills, line 179: I think this needs to go on the task execution role instead of the task role now. The task role is what is used at runtime, where as ECS assumes the execution role. We need ECS to fetch the values to inject into env vars

Mjwillis: ah ok got it

terraform/modules/data-share-service/iam.tf
```diff
@@ -1,6 +1,33 @@
 resource "aws_iam_role" "ecs_execution" {
   name               = "${var.environment}-execution-role"
   assume_role_policy = data.aws_iam_policy_document.assume_role_policy.json
+
+  inline_policy {
```
oskarwilliams378, line 5: Could we make this its own policy rather than an inline please :)

Mjwillis: Yes I just can't work out how to do that!

oskarwilliams378 (approved):

PR #439, opened 2023-02-07
GPC-93: Attempt deploy hook

oskarwilliams378:


terraform/modules/data-share-service/lb.tf
```diff
@@ -33,6 +33,24 @@ resource "aws_lb_listener" "listener-http" {
   depends_on = [aws_lb_target_group.green]
 }
 
+#tfsec:ignore:aws-elb-http-not-used
+resource "aws_lb_listener" "test-listener-http" {
```
olevett, line 37: Terraform encourages naming with _ as separators, but not too worried

terraform/modules/data-share-service/lb-security.tf
```diff
@@ -38,6 +38,23 @@ resource "aws_security_group" "lb_auto" {
   }
 }
 
+resource "aws_security_group" "lb_test" {
+  name_prefix = "${var.environment}-ecs-lb-test"
+  description = "Access to LB test port"
+  vpc_id      = module.vpc.vpc_id
+
+  ingress {
+    from_port   = 8080
```
olevett, line 47: Is there any risk here by exposing port 8080? Does that mean that clients can potentially reach this/a not-quite-production environment?

terraform/modules/deploy_hook/lambdas/hook.py
```diff
@@ -12,8 +12,8 @@
 logger = logging.getLogger()
 logger.setLevel(logging.INFO)
 
-gdx_url = os.environ["gdx_url"]
-events_url = gdx_url + "/events"
+test_gdx_url = os.environ["test_gdx_url"]
```
olevett, line 15: Should we expand "test" here, it's a smoke test specifically?

olevett (approved):

terraform/modules/data-share-service/lb.tf
oskarwilliams378: doh! will fix of course

terraform/modules/data-share-service/lb-security.tf
oskarwilliams378: Good point, I will look at limiting what can access this

terraform/modules/deploy_hook/lambdas/hook.py
oskarwilliams378: not really sure we should, as this is what codedeploy calls the "test" connection. Are you thinking we will get confused with it being a test env url vs a test url for this env?

terraform/modules/data-share-service/deploy_hook.tf
```diff
@@ -5,8 +5,29 @@ module "deploy_hook" {
...
+
+resource "aws_security_group" "lambda" {
+  name_prefix = "${var.environment}-ecs-lb-test"
+  description = "Access to vpc for lambda"
+  vpc_id      = module.vpc.vpc_id
+
+  egress {
+    from_port   = 0
+    to_port     = 0
+    protocol    = "-1"
```
ethanmills, line 25: Could this be http over 8080?

terraform/modules/deploy_hook/lambdas/hook.py
olevett: I was worried about it in the abstract as a "what does the word test even mean", if it's consistent with CodeDeploy and naming across the AWS side of things it's reasonable to keep it as is

terraform/modules/data-share-service/deploy_hook.tf
```diff
@@ -5,8 +5,29 @@ module "deploy_hook" {
...
+    description = "Allow access from lambda"
```
ethanmills, line 26: Is this not `Allow deployment hook lambda to call load balancer`?

terraform/modules/data-share-service/lb-security.tf
```diff
@@ -38,6 +38,24 @@ resource "aws_security_group" "lb_auto" {
...
+    to_port     = 8080
+    protocol    = "tcp"
+    description = "Allow access to LB from port 8080"
+    cidr_blocks = [var.vpc_cidr]
```
ethanmills, line 51: Could we use the lambda sg here instead?

terraform/modules/deploy_hook/lambdas/hook.py
```diff
@@ -38,6 +38,7 @@ def lambda_handler(event, _context):
         put_lifecycle_event_hook(event, 'Succeeded')
     except:
         put_lifecycle_event_hook(event, 'Succeeded')
```
ethanmills, line 40: Are we changing this back to Failed in another step?

terraform/modules/deploy_hook/lambdas/hook.py
oskarwilliams378: it is, CodeDeploy call it the test listener

ethanmills (approved):

terraform/modules/data-share-service/deploy_hook.tf
```diff
@@ -5,8 +5,29 @@ module "deploy_hook" {
...
+  security_group_id = aws_security_group.lambda.id
+  subnet_ids        = module.vpc.public_subnet_ids
+
+  test_gdx_url  = "http://${aws_lb.load_balancer.dns_name}:8080"
   auth_url      = module.cognito.token_auth_url
   client_id     = module.cognito.deploy_hook_client_id
   client_secret = module.cognito.deploy_hook_client_secret
 }
+
+resource "aws_security_group" "lambda" {
```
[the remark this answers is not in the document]
oskarwilliams378: Solved

terraform/modules/deploy_hook/lambdas/hook.py
oskarwilliams378: Will change to failed in a later commit

PR #443, opened 2023-02-07
GPC-69: Update LocalStack container version

olevett:


olevett:
Tests pass on this pr, and running locally with the same version (running, or tests) seemed good to me

PR #449, opened 2023-02-08
GPC-NONE: move deploy hook lambdas to private subnet

ethanmills:
This is fine and can access everything in the public subnet. However, the lb has no fixed private IP, so we cannot call it internal to the vpc. We'll need to allow public access on 8080 and do header based auth I think


ethanmills:
With the current state of the lambda, this is enough to get deployments going

oskarwilliams378 (approved):

PR #452, opened 2023-02-08
GPC-NONE: Align Gender to HMPPS source data

olevett:


oskarwilliams378 (approved):

ethanmills:
I thought the LEV API was sex?

ethanmills (changes requested):
Is this not a breaking API change?

olevett:
LEV _is_ Sex, and I haven't changed that, HMPPS is Gender - do we think changing HMPPS only is a breaking change?

ethanmills (approved):
Sorry, HMPPS and HMPO look too similar

oskarwilliams378 (commented):
Should we add a migration updating whatever is in the DB (for dev and demo), or just do that manually?

src/main/kotlin/uk/gov/gdx/datashare/enums/EnrichmentField.kt
```diff
@@ -20,4 +20,5 @@ enum class EnrichmentField(@JsonValue val jsonName: String) {
   PRISONER_NUMBER("prisonerNumber"),
   FIRST_NAME("firstName"),
   MIDDLE_NAMES("middleNames"),
+  GENDER("gender"),
```
Mjwillis, line 23: Why do we have 2 types of sex/gender ?  Is this because the two models have different results?

Mjwillis (commented):
Just not sure we need both but happy to approve if there is a good reason

src/main/kotlin/uk/gov/gdx/datashare/enums/EnrichmentField.kt
olevett: The reasoning is that they're fundamentally different things, that happen to have the same shape. I think we need to be very clear on them being different to our downstream acquirers.

PR #453, opened 2023-02-08
GPC-NONE: Prevent deploy hook from calling gdx

oskarwilliams378:


terraform/modules/data-share-service/deploy_hook.tf
```diff
@@ -3,7 +3,7 @@ module "deploy_hook" {
   environment                 = var.environment
   region                      = var.region
   cloudwatch_retention_period = var.cloudwatch_retention_period
-  codedeploy_arn              = aws_codedeploy_app.gdx_data_share_poc.arn
+  codedeploy_arn              = aws_codedeploy_deployment_group.gdx_data_share_poc.arn
```
ethanmills, line 6: What's this for?

ethanmills (approved):

terraform/modules/data-share-service/deploy_hook.tf
oskarwilliams378: Turns out it was wrong before, this is the correct arn

PR #455, opened 2023-02-08
GPC-NONE: Update cloudfront sg rule to use managed aws prefix list

oskarwilliams378:


terraform/modules/data-share-service/lb-security.tf
```diff
@@ -1,41 +1,35 @@
...
+  description       = "LB ingress rule for cloudfront"
+  security_group_id = aws_security_group.lb_cloudfront.id
+  prefix_list_ids   = [data.aws_ec2_managed_prefix_list.cloudfront.id]
+}
+
+resource "aws_security_group_rule" "lb_egress" {
+  type              = "egress"
+  protocol          = "-1"
+  from_port         = 0
+  to_port           = 0
```
ethanmills, line 29: Is there any reason this isn't more specific?

ethanmills (approved):

terraform/modules/data-share-service/lb-security.tf
oskarwilliams378: This is to make sure there are no extra changes here

PR #456, opened 2023-02-08
GPC-NONE: Restructure event handling

olevett:
Not sure how everyone feels about this as an approach, but it means adding an event handler is a little easier/doesn't need all the code changes.

If people like it, I'd want to rejig the tests
- each and every event type has exactly one handler
- splitting the aggregation from the event handlers themselves
- testing the aggregation


Mjwillis (approved):
Yes this is pretty much what I wanted to do when I added the PRISON event, but left it due to multiple commits going on at the time!  Great !

PR #457, opened 2023-02-08
GPC-NONE: Prefer single over first to enforce single handler

olevett:


Mjwillis (approved):
LGTM

olevett:
There's a question the failing tests make a little interesting, which is should we soft fail if there's no matching enrichment service or hard fail? Previously, we logged an error, which is what I've effectively done, but that relies on a "Null" reporter which can handle _any_ message and just no-ops+logs. For `.single` to work we can't have that no-op, because it means that every action has two possible consumers

PR #463, opened 2023-02-08
GPC-93: Allow access through cloudfront

oskarwilliams378:


terraform/modules/data-share-service/deploy_hook.tf
```diff
@@ -5,43 +5,9 @@ module "deploy_hook" {
...
-
-  test_gdx_url  = "http://${aws_lb.load_balancer.dns_name}:8080"
-  auth_url      = module.cognito.token_auth_url
-  client_id     = module.cognito.deploy_hook_client_id
-  client_secret = module.cognito.deploy_hook_client_secret
+  test_gdx_url     = "${local.gdx_api_base_url}:8443"
+  test_auth_header = random_password.test_auth_header.result
+  auth_url         = module.cognito.token_auth_url
+  client_id        = module.cognito.deploy_hook_client_id
+  client_secret    = module.cognito.deploy_hook_client_secret
```
ethanmills, line 12: Just to check, this change now means it doesn't access our VPC at all, so is back to normal lambda internet access?

terraform/modules/deploy_hook/variables.tf
```diff
@@ -3,12 +3,8 @@ variable "region" {}
 variable "cloudwatch_retention_period" {}
 variable "codedeploy_arn" {}
 
-variable "security_group_id" {}
-variable "subnet_ids" {
-  type = list(string)
-}
-
 variable "test_gdx_url" {}
+variable "test_auth_header" {}
```
ethanmills, line 7: Does this need to be marked as sensitive?

ethanmills (approved):

terraform/modules/data-share-service/deploy_hook.tf
oskarwilliams378: Exactly

PR #464, opened 2023-02-08
GPC-176 Add source ID as an enrichment field

Mjwillis:
Hide sourceId from event data unless specified (PII)


src/main/kotlin/uk/gov/gdx/datashare/services/EventDataService.kt
```diff
@@ -133,9 +133,11 @@ class EventDataService(
     return EventNotification(
       eventId = event.id,
       eventType = subscription.eventType,
-      sourceId = event.dataId,
+      eventTime = event.eventTime,
+      ingestTime = event.whenCreated,
+      sourceId = if (includeData && EnrichmentField.SOURCE_ID in enrichmentFieldNames) event.dataId else null,
       dataIncluded = if (!callbackEvent) includeData else null,
-      enrichmentFields = if (!callbackEvent) enrichmentFieldNames else null,
+      enrichmentFields = enrichmentFieldNames,
```
Mjwillis, line 140: Think we should always return these

src/main/resources/db/devonly/R__dev_data.sql
```diff
@@ -76,14 +76,14 @@ VALUES ('dwp-event-receiver', getIdFromAcquirerName('DWP Poller'),'DEATH_NOTIFIC
 
 INSERT INTO acquirer_subscription_enrichment_field(acquirer_subscription_id, enrichment_field)
 SELECT id,
-       unnest(ARRAY ['REGISTRATION_DATE', 'FIRST_NAMES', 'LAST_NAME', 'MAIDEN_NAME', 'DATE_OF_DEATH', 'DATE_OF_BIRTH', 'SEX', 'ADDRESS', 'BIRTH_PLACE', 'DEATH_PLACE', 'OCCUPATION', 'RETIRED'])
+       unnest(ARRAY ['SOURCE_ID', 'REGISTRATION_DATE', 'FIRST_NAMES', 'LAST_NAME', 'MAIDEN_NAME', 'DATE_OF_DEATH', 'DATE_OF_BIRTH', 'SEX', 'ADDRESS', 'BIRTH_PLACE', 'DEATH_PLACE', 'OCCUPATION', 'RETIRED'])
```
Mjwillis, line 79: explicitly return source id

src/main/kotlin/uk/gov/gdx/datashare/services/EventDataService.kt
ethanmills: I'd quite like not to return these at all for `/events/{id}` where they're present, but can't figure out how

Mjwillis: yes I notice this - it shows null which is annoying

src/main/kotlin/uk/gov/gdx/datashare/enums/EnrichmentField.kt
```diff
@@ -3,6 +3,8 @@ package uk.gov.gdx.datashare.enums
 import com.fasterxml.jackson.annotation.JsonValue
 
 enum class EnrichmentField(@JsonValue val jsonName: String) {
+  SOURCE_ID("sourceId"),
```
ethanmills, line 6: That's an interesting point. Do we consider death source ID to be the same type as prison source ID?

src/main/kotlin/uk/gov/gdx/datashare/services/EventDataService.kt
```diff
@@ -133,9 +133,11 @@ class EventDataService(
     return EventNotification(
       eventId = event.id,
       eventType = subscription.eventType,
-      sourceId = event.dataId,
+      eventTime = event.eventTime,
+      ingestTime = event.whenCreated,
+      sourceId = if (includeData && EnrichmentField.SOURCE_ID in enrichmentFieldNames) event.dataId else null,
```
oskarwilliams378, line 138: I think this means we would need to have 2 sets of enrichment fields, one for the "thin" and one for the "thick". This is because someone like DWP will want the source ID on the "thin" call as well

src/main/kotlin/uk/gov/gdx/datashare/enums/EnrichmentField.kt
Mjwillis: yes as its the same in the POST event (supplier) payload

src/main/kotlin/uk/gov/gdx/datashare/models/EventNotification.kt
```diff
@@ -22,8 +23,15 @@ data class EventNotification(
     example = "DEATH_NOTIFICATION",
   )
   val eventType: EventType,
-  @Schema(description = "ID from the source of the notification", required = true, example = "999999901")
-  val sourceId: String,
+
```
oskarwilliams378, line 26: We don't have this whitespace between other properties, I think we should be consistent but I don't care which way :)

src/main/resources/db/migration/V1_29__add_source_id_to_dwp_enrichment_fields.sql
```diff
@@ -0,0 +1,6 @@
+INSERT INTO acquirer_subscription_enrichment_field (acquirer_subscription_id, enrichment_field)
+SELECT asub.id, 'SOURCE_ID'
+FROM acquirer_subscription asub join acquirer a on asub.acquirer_id = a.id
+WHERE asub.event_type = 'DEATH_NOTIFICATION' and a.name like '%DWP%' and not exists (
+        select 1 from acquirer_subscription_enrichment_field asube where asube.acquirer_subscription_id = asub.id and asube.enrichment_field = 'SOURCE_ID'
+    );
```
oskarwilliams378, line 6: So I am not sure if it is best for this or to just manually add it in, happy for whichever but worth thinking that we are gonna flatten these migrations in a bit :)

oskarwilliams378 (changes requested):

src/main/kotlin/uk/gov/gdx/datashare/models/EventNotification.kt
Mjwillis: done

src/main/resources/db/migration/V1_29__add_source_id_to_dwp_enrichment_fields.sql
Mjwillis: I' leave it in for now

oskarwilliams378 (approved):

PR #467, opened 2023-02-09
GPC-60: Springboot3 upgrade

ethanmills:


ethanmills:
I need to do a bit more testing on this to make sure it hasn't broken anything

olevett (approved):

PR #468, opened 2023-02-09
GPC-177: Add initial autoapprove checks

oskarwilliams378:


ethanmills:
Is there a reason one is called autopass and the others are called autoapproved?

ethanmills (approved):

PR #471, opened 2023-02-09
GPC-93: Make the test lb listener public  not through cloudfront

oskarwilliams378:


ethanmills (approved):
Happy with this for now. Something I've just thought of is that if we put the lambda back in our VPC we can allow-list the EIP of the NAT gateway rather than making it public

oskarwilliams378:
I agree with get it working first!

PR #489, opened 2023-02-10
GPC-NONE: Remove Mockito dependency and update tests

olevett:


ethanmills (approved):

src/test/kotlin/uk/gov/gdx/datashare/queue/AwsQueueFactoryTest.kt
```diff
@@ -215,10 +241,12 @@ class AwsQueueFactoryTest {
...
       awsQueues = awsQueueFactory.createAwsQueues(sqsProperties)
 
-      verify(sqsClient).createQueue(
-        check<CreateQueueRequest> {
-          assertThat(it.attributes()).containsEntry(QueueAttributeName.REDRIVE_POLICY, """{"deadLetterTargetArn":"some dlq arn","maxReceiveCount":"2"}""")
-        },
+      val createQueueRequest = mutableListOf<CreateQueueRequest>()
+      verify { sqsClient.createQueue(capture(createQueueRequest)) }
+
+      assertThat(createQueueRequest.last().attributes()).containsEntry(
```
olevett, line 247: There's an odd interaction here that I think's worth calling out - the  `awsQueueFactory...` method gets called in a `BeforeEach`  as well as in this method, which means this gets called twice but we only care about the last one

PR #490, opened 2023-02-10
GPC-NONE: Add in trivy

oskarwilliams378:


.trivyignore
```diff
@@ -0,0 +1,6 @@
+## DOCKER
+
+# ADD is used for getting file from the internet
+AVD-DS-0005
+# Healthcheck already exists
+AVD-DS-0026
```
olevett, line 6: I don't think is is true - we have a healthcheck in the docker-compose and in ECS config, but we don't have one in our dockerfile itself

olevett:
What gap is this filling out of interest, is it earlier/different coverage to e.g. ECS scanning?

pyproject.toml
```diff
@@ -1,17 +0,0 @@
-[tool.poetry]
```
olevett, line 1: Is this not used by the lambdas?

.trivyignore
oskarwilliams378: Will update

oskarwilliams378:
This will fill the gap of ECR image scanning, I'll add a daily job for this afterwards. It also allows us to remove snyk. Basically it gets us down to one system in one place which is quite nice!

olevett (approved):

PR #505, opened 2023-02-14
GPC-146 alarms for error rates

CarlyG55:


terraform/modules/alarm/main.tf
```diff
@@ -0,0 +1,41 @@
+resource "aws_cloudwatch_metric_alarm" "error_rate_alarm" {
+  alarm_name          = var.alarm_name
+  comparison_operator = "GreaterThanOrEqualToThreshold"
+  evaluation_periods  = "2"
+  threshold           = "10"
+  alarm_description   = var.alarm_description
+  treat_missing_data  = "notBreaching"
+  alarm_actions       = [var.alarm_action]
+  ok_actions          = [var.alarm_action]
```
ethanmills, line 9: Are these two meant to be the same?

terraform/modules/alarm/main.tf
```diff
@@ -0,0 +1,41 @@
...
+
+  metric_query {
+    id          = "error_rate"
+    expression  = "IF(successful_requests + errors > 10, 100*errors/(successful_requests + errors), 0)"
```
oskarwilliams378, line 13: maybe put some spacing around operators here :)

terraform/modules/data-share-service/alarms.tf
```diff
@@ -0,0 +1,179 @@
...
+        namespace = local.metric_namespace,
+        dimensions = {
+          success = true
+        }
+      },
+    },
+  }
+}
+
+module "error_rate_alarm" {
```
oskarwilliams378, line 103: Possibly make this the below as this module will generate many
```suggestion
module "error_rate_alarms" {
```

terraform/modules/data-share-service/alarms.tf
```diff
@@ -0,0 +1,179 @@
+locals {
+  alarm_prefix = "${var.environment}-gdx"
+  alarms = {
+    lev_error_rate = {
+      alarm_name        = "${local.alarm_prefix}-lev-error-rate",
```
oskarwilliams378, line 5: We should probably add the alarm actions to these, posting to the SNS, just it won't email

terraform/modules/data-share-service/outputs.tf
```diff
@@ -16,3 +16,6 @@ output "consumer_client_id" {
 output "consumer_client_secret" {
   value = module.cognito.consumer_client_secret
 }
+output "alarm_action" {
```
oskarwilliams378, line 19: why is this being outputted?

oskarwilliams378 (changes requested):

terraform/modules/data-share-service/alarms.tf
```diff
@@ -0,0 +1,179 @@
...
+      alarm_description = "LEV error rate",
+      error_metric = {
+        name      = "API_RESPONSES.ResponsesFromLev",
+        namespace = local.metric_namespace,
+      },
+      success_metric = {
+        name      = "API_RESPONSES.ErrorsFromLev",
+        namespace = local.metric_namespace,
```
ethanmills, line 13: Are these two the right way round?

terraform/modules/data-share-service/alarms.tf
```diff
@@ -0,0 +1,179 @@
...
+      alarm_name        = "${local.alarm_prefix}-prisoner-search-error-rate",
+      alarm_description = "Prisoner search error rate",
+      error_metric = {
+        name      = "API_RESPONSES.ResponsesFromPrisonerSearch",
+        namespace = local.metric_namespace,
+      },
+      success_metric = {
+        name      = "API_RESPONSES.ErrorsFromPrisonerSearch",
+        namespace = local.metric_namespace,
+      },
```
ethanmills, line 26: ditto

terraform/modules/data-share-service/sns_alarm.tf
```diff
@@ -0,0 +1,34 @@
...
+        }
+      },
+      "Outputs" : {
+        "ARN" : {
+          "Description" : "Email SNS Topic ARN",
+          "Value" : { "Ref" : "EmailSNSTopic" }
+        }
+      }
+    }
+  )
```
ethanmills, line 33: Are we using this because it's easier than setting up a topic and topic subscription in tf?

oskarwilliams378: This is from me, because of the old situation where you couldn't at all sub an sns topic using terraform via email, however it looks possible now looking at https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/sns_topic_subscription

terraform/modules/alarm/main.tf
oskarwilliams378: Yep, these just tell you what system to notify, so we want it to notify the SNS topic, but it will notify it differently depending on it being alarm or ok

terraform/modules/data-share-service/outputs.tf
CarlyG55: oh yep shouldn't be - this was how I originally did it

terraform/modules/data-share-service/alarms.tf
oskarwilliams378: Am dumb, seen this below

terraform/modules/alarm/main.tf
CarlyG55: This was on @oskarwilliams378's example from lner

terraform/modules/data-share-service/sns_alarm.tf
CarlyG55: Have given this a go!

oskarwilliams378 (approved):

ethanmills (approved):

PR #514, opened 2023-02-15
GPC-181: Add timed and counted for events

oskarwilliams378:


src/main/kotlin/uk/gov/gdx/datashare/controllers/EventsController.kt
```diff
@@ -184,6 +183,8 @@ class EventsController(
       ),
     ],
   )
+  @Timed("time.eventsController.getEvents", percentiles = [0.5, 0.90, 0.99])
+  @Counted("count.eventsController.getEvents")
```
ethanmills, line 187: What does this do that we don't get already?

src/main/kotlin/uk/gov/gdx/datashare/config/MicrometerConfiguration.kt
```diff
@@ -24,17 +26,22 @@ class MicrometerConfiguration {
...
-    val cloudWatchConfig: CloudWatchConfig = setupCloudWatchConfig(namespace)
-    return CloudWatchMeterRegistry(
-      cloudWatchConfig,
-      Clock.SYSTEM,
-      cloudWatchAsyncClient(),
-    )
-  }
+
+  @Bean
+  fun timedAspect(registry: MeterRegistry) = TimedAspect(registry)
```
ethanmills, line 41: Why do we need this? Spring times controllers by default I thought: 
https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java#L49
https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#actuator.metrics.supported.spring-mvc

src/main/kotlin/uk/gov/gdx/datashare/controllers/EventsController.kt
oskarwilliams378: My plan is to move to this style of doing our count metrics instead of using the meter registries and incrementing etc as we will be using the `@Timed` annotation anyways.

terraform/modules/data-share-service/cloudwatch.tf
```diff
@@ -41,23 +41,43 @@ module "metrics_dashboard" {
...
+            exception = "none",
+            method    = "publishEvent",
+            class     = "uk.gov.gdx.datashare.controllers.EventsController"
+          },
           attributes = { label = "PublishEvent" }
         },
         {
-          name       = "API_CALLS.GetEvent.count",
-          dimensions = { success = true },
+          name = "count.eventsController.getEvent.count",
```
ethanmills, line 54: Is this different from `http.server.requests.count` where `exception = 'none' and method = 'GET' and uri = '/events' and outcome = 'SUCCESS'`?

src/main/kotlin/uk/gov/gdx/datashare/config/MicrometerConfiguration.kt
oskarwilliams378: We need this for our percentiles. The other approach would be doing a meter filter. However that wouldn't then help with this case as micrometer times each and doesn't detect it is just an `/{id}`
image

terraform/modules/data-share-service/cloudwatch.tf
```diff
@@ -41,23 +41,43 @@ module "metrics_dashboard" {
...
+            exception = "none",
+            method    = "getEvent",
+            class     = "uk.gov.gdx.datashare.controllers.EventsController"
+          },
           attributes = { label = "GetEvent" }
         },
         {
-          name       = "API_CALLS.GetEvents.count",
-          dimensions = { success = true },
+          name = "count.eventsController.getEvents.count",
```
ethanmills, line 64: Is this different from http.server.requests.count where exception = 'none' and method = 'GET' and uri = '/events' and outcome = 'SUCCESS'? Or are we counting the number of individual events somehow?

oskarwilliams378: Nope there is no difference here, my thought is that we want a consistent way of generating our metrics, and that we don't want to grab some from the automagic ones, and some where we define them

src/main/kotlin/uk/gov/gdx/datashare/config/MicrometerConfiguration.kt
oskarwilliams378: For percentiles I have found a work around, but it doesn't help with the image above I think

terraform/modules/data-share-service/cloudwatch.tf
ethanmills: This is still automagic, just in a worse way IMO

src/main/kotlin/uk/gov/gdx/datashare/config/MicrometerConfiguration.kt
ethanmills: That's for our client calls to the LEV, which this doesn't affect. You can do `http.server.requests.count` with `{id}` just fine. If we want to collect aggregate stats on our client calls I'd rather tag the requests and use the default metrics I think

terraform/modules/data-share-service/cloudwatch.tf
oskarwilliams378: Sure

src/main/kotlin/uk/gov/gdx/datashare/config/MicrometerConfiguration.kt
oskarwilliams378: Sure

src/main/kotlin/uk/gov/gdx/datashare/controllers/EventsController.kt
oskarwilliams378: Removing @TImed

src/main/resources/application.yml
```diff
@@ -86,6 +86,10 @@ management:
     info:
       cache:
         time-to-live: 2000ms
+  metrics:
+    distribution:
+      percentiles:
+        http.server.requests: 0.9,0.99
```
olevett, line 92: Do we get a mean/median time by default? I think whilst the 99% centile is generally better for limits, understanding what a middle of the road consumer gets can be very ehlpful

olevett (approved):

ethanmills (approved):

src/main/resources/application.yml
ethanmills: Why is prometheus so much better than cloudwatch? https://micrometer.io/docs/concepts#_histograms_and_percentiles

oskarwilliams378:
Removed from queue as the rebase has shown a load of issues

src/main/resources/application.yml
oskarwilliams378: We get mean by default

olevett (approved):

PR #530, opened 2023-02-17
GPC-185: Create baseline migration

CarlyG55:


src/main/resources/db/baseline/B1_29__baseline.sql
```diff
@@ -0,0 +1,86 @@
+CREATE TABLE supplier
```
ethanmills, line 1: Do the B prefixes work? I thought you needed to pay for flyway enterprise to get that

src/main/resources/db/devonly/R__dev_data.sql
```diff
@@ -30,36 +30,14 @@ Begin
 End;
 $$;
 
-DELETE
```
oskarwilliams378, line 33: These are to make the migration repeatable, everything added should be removed

src/main/resources/db/migration/V1__baseline.sql
```diff
@@ -0,0 +1,86 @@
...
+    id                       UUID PRIMARY KEY         NOT NULL DEFAULT gen_random_uuid(),
+    acquirer_subscription_id UUID                     NOT NULL,
+    data_id                  VARCHAR(80)              NOT NULL,
+    when_created             TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
+    event_time               TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
+    deleted_at               TIMESTAMP WITH TIME ZONE,
+
+    CONSTRAINT fk_acquirer_subscription
+        FOREIGN KEY (acquirer_subscription_id)
+            REFERENCES acquirer_subscription
```
oskarwilliams378, line 56: does this need this?
```suggestion
            REFERENCES acquirer_subscription (id)
```

src/main/resources/db/migration/V1__baseline.sql
```diff
@@ -0,0 +1,86 @@
+CREATE TABLE supplier
```
oskarwilliams378, line 1: could we sort these creations a little more, either by type or by alphabet (mainly thinking the `acquirer_subscription_enrichment_field` should be after the `acquirer_subscription`)

oskarwilliams378 (changes requested):

oskarwilliams378 (approved):

PR #538, opened 2023-02-20
GPC-191: Fix prometheus remote write

oskarwilliams378:


prometheus/Adot.Dockerfile
```diff
@@ -1,4 +1,30 @@
-FROM public.ecr.aws/aws-observability/aws-otel-collector:v0.26.1
+FROM public.ecr.aws/aws-observability/aws-otel-collector:latest as base
```
ethanmills, line 1: Was this needed?

prometheus/Adot.Dockerfile
```diff
@@ -1,4 +1,30 @@
...
+
+FROM alpine:latest as run
```
ethanmills, line 3: can we pin this?

ethanmills (approved):

PR #546, opened 2023-02-21
GPC-NONE: Update to gradle 8

ethanmills:


olevett:
Presumably we need to upgrade the docker base image and or the github actions that reference gradle directly?

ethanmills:
I think this is going to cause more trouble than it's worth for the time being. I'll wait until spring have had a chance to catch up and officially support gradle 8

PR #550, opened 2023-02-21
GPC-NONE: Add cognito to grafana

oskarwilliams378:


terraform/shared/modules/grafana/cognito.tf
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  software_token_mfa_configuration {
+    enabled = true
+  }
+
+  verification_message_template {
+    default_email_option = "CONFIRM_WITH_LINK"
+    email_subject        = "Verify access to Grafana"
+    email_message        = "Click the link below to verify access to Grafana \n {####}"
```
ethanmills, line 22: I assume this is some mysterious magic incantation?

ethanmills (approved):

terraform/shared/modules/grafana/cognito.tf
oskarwilliams378: Ooogily Boogily now you're in cognito

PR #552, opened 2023-02-22
GPC-194: Setup prometheus alerts manager

oskarwilliams378:


terraform/modules/data-share-service/prometheus.tf
```diff
@@ -11,11 +11,47 @@ resource "random_password" "prometheus_password" {
...
+      "aps:GetMetricMetadata",
+      "aps:GetSeries",
+      "aps:QueryMetrics",
+      "aps:ListRules",
+      "aps:ListAlertManagerSilences",
+      "aps:ListAlertManagerAlerts",
+      "aps:GetAlertManagerStatus",
+      "aps:ListAlertManagerAlertGroups",
+      "aps:PutAlertManagerSilences",
+      "aps:DeleteAlertManagerSilences",
```
ethanmills, line 27: Does this mean that we _can_ set up AMP alert manager as a datasource in grafana?

terraform/modules/data-share-service/prometheus.tf
```diff
@@ -11,11 +11,47 @@ resource "random_password" "prometheus_password" {
...
+resource "aws_prometheus_alert_manager_definition" "prometheus" {
+  workspace_id = aws_prometheus_workspace.prometheus.id
+  definition   = <<EOF
+alertmanager_config: |
+  route:
+    receiver: 'default'
+  receivers:
+    - name: 'default'
+      sns_configs:
+      - topic_arn: ${aws_sns_topic.sns_alarm_topic.arn}
```
ethanmills, line 53: Is this the topic we've subscribed the email address to?

ethanmills (approved):

terraform/modules/data-share-service/prometheus.tf
oskarwilliams378: Yep :)

terraform/modules/data-share-service/prometheus.tf
oskarwilliams378: Yep :)

PR #553, opened 2023-02-22
GPC-151: Use corretto for Java distribution

olevett:
Part 1 of GPC-151 is to use Corretto, I want to check this works nicely / boots up properly, then will try playing with SSL config

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM eclipse-temurin:19-jre-jammy AS builder
+FROM amazoncorretto:19-alpine AS builder
 
 WORKDIR /app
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -13,7 +13,7 @@ ENV BUILD_NUMBER ${BUILD_NUMBER:-1_0_0}
 COPY . .
 RUN ./gradlew clean assemble -Dorg.gradle.daemon=false
 
-FROM eclipse-temurin:19-jre-alpine
```
ethanmills, line 16: Is it worth building with corretto too? (line 1)

ethanmills (approved):

Dockerfile
olevett: Will see if it works locally and if so, do that

PR #559, opened 2023-02-22
GPC-151: Attempt verify-full and installing the cert 

olevett:


oskarwilliams378 (approved):

ethanmills (approved):
It may be the I'm not checking correctly, but it seems the rds cert isn't in the amazon-cacerts bundled with the jre 😞 https://github.com/corretto/corretto-jdk/blob/develop/amazon-cacerts

PR #560, opened 2023-02-22
GPC-NONE: Prometheus logs and base rule

oskarwilliams378:


terraform/modules/data-share-service/prometheus.tf
```diff
@@ -55,3 +59,18 @@ alertmanager_config: |
           region: eu-west-2
 EOF
 }
+
+resource "aws_prometheus_rule_group_namespace" "test" {
```
ethanmills, line 63: Is this intended to stay longterm?

ethanmills (approved):

terraform/modules/data-share-service/prometheus.tf
oskarwilliams378: Nope, just with all the chaos right now with merge and infra I want this in our codebase so that I can then remove with a cleanup PR

PR #561, opened 2023-02-22
GPC-NONE: Add prometheus sns permissions

oskarwilliams378:


terraform/modules/data-share-service/sns_alarm.tf
```diff
@@ -1,5 +1,7 @@
 locals {
   notification_emails = ["gdx-dev-team@digital.cabinet-office.gov.uk"]
+
+
```
olevett, line 4:
```suggestion
```

terraform/modules/data-share-service/sns_alarm.tf
```diff
@@ -17,6 +19,46 @@ resource "aws_sns_topic" "sns_alarm_topic" {
...
+    actions = [
+      "SNS:Subscribe",
+      "SNS:SetTopicAttributes",
+      "SNS:RemovePermission",
+      "SNS:Receive",
+      "SNS:Publish",
+      "SNS:ListSubscriptionsByTopic",
+      "SNS:GetTopicAttributes",
+      "SNS:DeleteTopic",
+      "SNS:AddPermission",
```
olevett, line 33: This seems like lots of access, but I'm not too familiar with SNS permissions

oskarwilliams378: These are the default SNS permissions, I can change it so that our prometheus only has a subset?

PR #562, opened 2023-02-22
Techdocs content

ethanmills:
There's still more that could be done here, but I'm going to leave it at this for now. I feel like enough will change before this gets used that it's  not worth putting too much polish on just yet

I've also not touched the supplier side, but it's templated out.


doc/how-to-onboard-an-acquirer.md
```diff
@@ -1,10 +1,10 @@
-# How to onboard a client
+# How to onboard an acquirer
```
ethanmills, line 1: This doc is now out of date anyway because of the api endpoint @oskarwilliams378 added

src/main/kotlin/uk/gov/gdx/datashare/controllers/EventsController.kt
```diff
@@ -49,43 +49,9 @@ class EventsController(
...
-      "        \"prisonerNumber\": \"A1234DB\",\n" +
-      "        \"firstName\": \"Mary\",\n" +
-      "        \"middleNames\": \"Jane\",\n" +
-      "        \"lastName\": \"Smith\",\n" +
-      "        \"gender\": \"Male\",\n" +
-      "        \"dateOfBirth\": \"1972-02-20\",\n" +
-      "}\n" +
-      "      </pre>" +
-      "      <p><b>Mandatory Fields</b>: prisonerNumber, firstName, lastName, gender, dateOfBirth</p>",
+    description = "<h3>Event Types</h3>" +
```
olevett, line 52: Do we want to remove the `enrichmentFieldsIncludedInPoll` option?

olevett (approved):

techdocs/source/index.html.md.erb
```diff
@@ -5,21 +5,17 @@ weight: 10
 
 # About the GDX Platform
 
-<%= warning_text("You can use the Technical Documentation Template (TDT) to create and publish technical documentation for users in government, including API documentation.<br/><br/>
-We’ve provided these starting pages with a best practice API documentation structure. It includes the template structure of the documentation sections you need, and is based on <a href='https://www.gov.uk/guidance/how-to-document-apis'>our latest guidance on how to document APIs</a>.<br/><br/>
-You can also start from scratch if you’re not using the TDT to create API documentation.") %>
+<%= warning_text("GDX is not a live service. All data is fake test data.") %>
```
olevett, line 8:
```suggestion
<%= warning_text("GDX is not a live service. All data is synthetic test data.") %>
```

src/main/kotlin/uk/gov/gdx/datashare/controllers/EventsController.kt
ethanmills: I think so. It looks like a performance nightmare

olevett: https://govukverify.atlassian.net/browse/GPC-197

PR #565, opened 2023-02-22
GPC-NONE: Add dashboard

oskarwilliams378:


grafana/Dashboards/Requests.json
```diff
@@ -0,0 +1,172 @@
+{
+  "__inputs": [
+    {
+      "name": "DS_PROMETHEUS_- DEV",
```
olevett, line 4: Do we want one for demo too? Probably not going to show much now, but might be good to figure out the pattern of "this same dashboard in two places"

olevett (approved):

PR #571, opened 2023-02-23
GPC-NONE: shuffle gauge configuration to persist value between calls

olevett:
I think this is the problem with the UnconsumedEvents count, where when Micrometer is gathering metrics, there's no value available (the gauge itself doesn't really hold the value)

### DIFF
--- a/src/main/kotlin/uk/gov/gdx/datashare/services/ScheduledJobService.kt
+++ b/src/main/kotlin/uk/gov/gdx/datashare/services/ScheduledJobService.kt
@@ -1,6 +1,5 @@
 package uk.gov.gdx.datashare.services
 
-import io.micrometer.core.instrument.Gauge
 import io.micrometer.core.instrument.MeterRegistry
 import net.javacrumbs.shedlock.core.LockAssert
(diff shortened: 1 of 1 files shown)


olevett:
FYI @oskarwilliams378

src/main/kotlin/uk/gov/gdx/datashare/services/ScheduledJobService.kt
```diff
@@ -1,25 +1,27 @@
...
 import java.util.concurrent.TimeUnit
+import java.util.concurrent.atomic.AtomicInteger
 
 @Service
 class ScheduledJobService(
   private val eventDataRepository: EventDataRepository,
-  private val meterRegistry: MeterRegistry,
+  meterRegistry: MeterRegistry,
 ) {
+  private val guage = meterRegistry.gauge("UnconsumedEvents", AtomicInteger(0))
```
ethanmills, line 17: spelling?

olevett: Eugh yes

oskarwilliams378 (approved):

PR #586, opened 2023-02-24
GPC-187: ADR for SQS

ethanmills:
For discussion

### DIFF
--- a/doc/architecture/decisions/0016-push-directly-to-sqs.md
+++ b/doc/architecture/decisions/0016-push-directly-to-sqs.md
@@ -1,0 +1,47 @@
+# 16. Push events directly to SQS
+[Next >>](9999-end.md)
+
+Date: 2023-02-23
+
+## Status
+
+Proposal for discussion
+
+## Context
+
+Some technically mature acquirers require a queue type system for events, rather than the polling API interface currently
+exposed. There are considerations around what the API boundary of the service is, and how we present events on a queue
+while still meeting audit logging and data retention requirements. Our guiding principles here are:
+- Protect sensitive data from unauthorised access
+- Minimise the number of components exposed to sensitive information
+- Maintain robust audit logging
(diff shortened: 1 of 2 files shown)


doc/architecture/decisions/0016-push-directly-to-sqs.md
```diff
@@ -0,0 +1,44 @@
...
+
+More broadly, the API surface of the system will be expanded to include the acquirer queue as well as the HTTP REST API.
+AWS services such as SQS and S3 may be considered for inclusion in the API surface exposed, whereas third party systems
+are more likely to need an adaptor. Authentication to AWS services can happen with ephemeral credentials created using IAM
+roles, but external services will need longer term credentials. It is also easier to monitor compliance with policies using
+built in AWS tooling. We should therefore consider anything external to AWS to outside of the "core" service and so
+requiring considerations.
+
+## Impact
+Exposing acquirer queues to the kotlin service will mean that queue clients need to be generated dynamically from data in
```
olevett, line 38: With this, do we also implicitly take on creating queues as well as clients from our service?

olevett:
I think I'm broadly happy with this as an approach and justification. I think as/when we've implemented this, it would be good to have a brief review and validate it is working as we think it should.

doc/architecture/decisions/0016-push-directly-to-sqs.md
ethanmills: Yes, I can make this explicit

oskarwilliams378:
I agree with Olly, I think this approach makes sense with out current expectation of the system

Mjwillis:
Yes sounds like a good starting point - suspect we'll come up with some interesting issue on route when we implement but I like to idea of taking away the complexity of setup on the acquirers side as much as we can.

Mjwillis (approved):
See comments

PR #588, opened 2023-02-24
GPC-NONE: Add histograms to all timers

oskarwilliams378:


src/main/resources/application.yml
```diff
@@ -92,6 +92,7 @@ management:
         http:
           server.requests: true
           client.requests: true
+          DATA_PROCESSING.TimeFromCreationToDeletion: true
```
olevett, line 95: I take it this is the only one that's not got histograms then?

olevett (approved):

src/main/resources/application.yml
oskarwilliams378: Yep

PR #590, opened 2023-02-24
GPC-196: Add data enrichment

oskarwilliams378:


ethanmills (approved):
As a general point, I really value screenshots of anything that can be seen on reviews. I find this really hard to parse without context

PR #597, opened 2023-02-27
GPC-NONE: commonise event details

ethanmills:
Just some small changes to add some types and make things a little more consistent. This will make my SQS work easier

### DIFF
--- a/src/main/kotlin/uk/gov/gdx/datashare/models/PrisonerDetails.kt
+++ b/src/main/kotlin/uk/gov/gdx/datashare/models/PrisonerDetails.kt
@@ -1,14 +1,20 @@
 package uk.gov.gdx.datashare.models
 
-import com.fasterxml.jackson.annotation.JsonInclude
+import com.fasterxml.jackson.core.JsonGenerator
+import com.fasterxml.jackson.databind.SerializerProvider
+import com.fasterxml.jackson.databind.annotation.JsonSerialize
+import com.fasterxml.jackson.databind.ser.std.StdSerializer
 import io.swagger.v3.oas.annotations.media.Schema
 import org.springframework.format.annotation.DateTimeFormat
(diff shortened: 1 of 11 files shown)


src/main/kotlin/uk/gov/gdx/datashare/models/PrisonerDetails.kt
```diff
@@ -28,4 +34,50 @@ data class PrisonerDetails(
   @Schema(description = "Date the prisoner was born", required = false, example = "2001-12-31", type = "date")
   @DateTimeFormat(iso = DateTimeFormat.ISO.DATE)
   val dateOfBirth: LocalDate? = null,
-)
+) : EventDetails {
+  companion object : EventDetailsCompanionObject<PrisonerDetails, PrisonerRecord> {
```
olevett, line 38: I was pondering this recently, but we do the filtering of fields twice, if we only ever serialize to Json using the serializer in this file, I wonder if we can simplify that at all, but that's possibly a step too far

olevett (approved):

src/main/kotlin/uk/gov/gdx/datashare/models/PrisonerDetails.kt
ethanmills: Yes, I think I might need to do some tweaking of this for the queue work anyway. If we have two queue consumers we'll end up calling the upstream twice at processing time at the moment. Not the end of the world, but a bit of a smell

olevett: I was trying to work out if there's a generic "field" type that does the combination of these things, and strips out some of the work for us - the more times we repeat the filtering code, the more I'm worried we might have one subtle difference and end up feeding data through we shouldn't!

PR #600, opened 2023-02-28
GPC-184: Document mTLS options

olevett:


doc/architecture/decisions/0017-mutual-tls.md
```diff
@@ -0,0 +1,46 @@
...
+We can potentially handle mTLS in the Spring/Kotlin service layer, but this adds significant complexity.
+Alternatively, AWS API Gateway offers management of mTLS.
+
+API Gateway doesn't support CRLs natively, which might need a custom lambda to manage, additionally
+we need a way to feed back the nature of the certificate to the upstream service, see AWS [blog](
+https://aws.amazon.com/blogs/compute/propagating-valid-mtls-client-certificate-identity-to-downstream-services-using-amazon-api-gateway/).
+
+## Decision
+
+For now, don't implement mTLS, when appropriate we should
```
ethanmills, line 30: I think I agree with the implementation here, although it might be worth exploring what the other options are (do cloudflare make it trivial for example, and would it be worth adding another layer to achieve that). What I'm less clear on is what mTLS brings over and above our existing authentication.

doc/architecture/decisions/0017-mutual-tls.md
```diff
@@ -0,0 +1,46 @@
...
+Date: 2023-02-28
+
+## Status
+
+Proposal for discussion
+
+## Context
+
+Some acquirers and data suppliers have a desire for additional security measures over and above encryption in transit
+and the OIDC authentication currently in place, as part of a Zero Trust approach.
```
ethanmills, line 13: I'm not sure we can currently claim a Zero Trust approach. Perhaps we should, but then mTLS termination at the gateway might not be a good fit - we would certainly need some encryption in transit over the VPC

olevett: I think I was using Zero Trust here as a "it's possible departments are moving towards this so would want all their touch points to do so"

ethanmills (approved):

PR #618, opened 2023-03-06
GPC-101 Acquirer management UI

CarlyG55:
A basic acquirer management that runs locally but is not authenticated or hosted

### DIFF
--- a/frontend/server/config.ts
+++ b/frontend/server/config.ts
@@ -1,6 +1,6 @@
 const production = process.env.NODE_ENV === 'production'
 
-function get<T>(name: string, fallback: T, options = { requireInProduction: false }): T | string {
+function get<T>(name: string, fallback: T, options = {requireInProduction: false}): T | string {
     if (process.env[name]) {
         return process.env[name]
     }
@@ -10,7 +10,7 @@ function get<T>(name: string, fallback: T, options = { requireInProduction: fals
     throw new Error(`Missing env var ${name}`)
 }
 
-const requiredInProduction = { requireInProduction: true }
+const requiredInProduction = {requireInProduction: true}
 
 export class AgentConfig {
     timeout: number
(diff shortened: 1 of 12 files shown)


frontend/server/config.ts
```diff
@@ -0,0 +1,48 @@
...
+        deadline: number
+    }
+    agent: AgentConfig
+}
+
+export default {
+    staticResourceCacheDuration: 20,
+    apis: {
+        gdxDataSharePoc: {
+            url: get('GDX_DATA_SHARE_URL', 'http://127.0.0.1:63459', requiredInProduction),
```
CarlyG55, line 36: This is the port gdx-data-share-poc was running on when I last was running the site. Is there a way of pulling this through, or should I just add a comment?

oskarwilliams378: We should probably set this to 8080, because thats what it runs on in everything except for when running the docker compose through IDEA

frontend/server/data/restClient.ts
```diff
@@ -0,0 +1,79 @@
...
+                .set(headers)
+                .responseType(responseType)
+                .timeout(this.timeoutConfig())
+
+            return raw ? result : result.body
+        } catch (error) {
+            const sanitisedError = sanitiseError(error)
+            throw sanitisedError
+        }
+    }
```
oskarwilliams378, line 78: Just to deduplicate some code, how about something like 

```suggestion
    async get({path = null, query = '', headers = {}, responseType = '', raw = false}: GetRequest): Promise<unknown> {
        const request = superagent
            .get(`${this.apiUrl()}${path}`)
            .query(query)
        return await this.makeRequest(request, headers, responseType, raw)
    }

    async post(
        {
            path = null,
            headers = {},
            responseType = '',
            data = {},
            raw = false,
        }: PostRequest = {}): Promise<unknown> {
        const request = superagent
            .post(`${this.apiUrl()}${path}`)
            .send(data)
        return await this.makeRequest(request, headers, responseType, raw)
    }

    async makeRequest(
        request: superagent.SuperAgentRequest,
        headers = {},
        responseType: string,
        raw: boolean): Promise<superagent.Response | any> {
        try {
            const result = await request.agent(this.agent)
                .auth(this.token, {type: 'bearer'})
                .set(headers)
                .responseType(responseType)
                .timeout(this.timeoutConfig())

            return raw ? result : result.body
        } catch (error) {
            throw sanitiseError(error)
        }
    }
```

oskarwilliams378 (changes requested):

oskarwilliams378 (approved):

PR #621, opened 2023-03-06
GPC-211: Add multi region data backups

oskarwilliams378:


olevett:
This seems fine, I think I'd like to understand
- what we think this protects us from
- how we recover
In particular, does this help us if we have an AWS account level breach? Does this help if we release a dodgy change etc?

Can we document this as a runbook or similar please?

PR #651, opened 2023-03-10
GPC-148: Anomalous traffic alarms

CarlyG55:


terraform/modules/data-share-service/prometheus.tf
```diff
@@ -65,12 +86,9 @@ resource "aws_prometheus_rule_group_namespace" "alerts" {
...
-  - name: Events
+  - name: alerts
     rules:
-    - alert: Growing Unconsumed Events
-      expr: max(UnconsumedEvents) > 500000
-      for: 5m
-      annotations:
-        summary: Over 500000 unconsumed events in database
+${local.metric_rules}
+${local.alert_rules}
```
oskarwilliams378, line 92: Could we make this multiple group pls, so something like 
```
  - name: records
    rules:
${local.metric_rules}
  - name: alerts
    rules:
${local.alert_rules}
```

oskarwilliams378 (approved):

PR #652, opened 2023-03-10
GPC-NONE: Remove unused API endpoints

ethanmills:


oskarwilliams378 (approved):

src/main/kotlin/uk/gov/gdx/datashare/controllers/AdminController.kt
```diff
@@ -11,80 +10,17 @@ import org.springframework.http.MediaType
...
-    ],
-  )
-  fun createCognitoClient(
-    @Schema(
-      required = true,
-      implementation = CognitoClientRequest::class,
-    )
-    @RequestBody
-    cognitoClientRequest: CognitoClientRequest,
-  ) = adminService.createCognitoClient(cognitoClientRequest)
```
olevett, line 86: I assume we're still doing this as part of the "onboard an acquirer" workflows? If not, can we reduce some roles/generally tighten access?

ethanmills: Yeah, still doing it there, but we currently have no use case for doing it separately (this won't rotate creds)

PR #661, opened 2023-03-13
GPC-120: Add enhanced monitoring role

oskarwilliams378:


terraform/modules/data-share-service/rds.tf
```diff
@@ -34,6 +34,31 @@ resource "aws_rds_cluster" "rds_postgres_cluster" {
   }
 }
 
+data "aws_iam_policy_document" "rds_enhanced_monitoring_assume_role" {
```
ethanmills, line 37: What does enhanced monitoring give us?

oskarwilliams378: Gives us access to better monitoring of the actual DB that's running
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_Monitoring.OS.overview.html

ethanmills (approved):

PR #678, opened 2023-03-15
GPC-253: Setup hosted zones

oskarwilliams378:


terraform/shared/main.tf
```diff
@@ -90,3 +90,7 @@ module "github_iam" {
   environment = local.env
   account_id  = data.aws_caller_identity.current.account_id
 }
+
+resource "aws_route53_zone" "zone" {
+  name = "share-life-events.service.gov.uk"
```
ethanmills, line 95: I think there's probably a step missing here to delegate the subdomains down

oskarwilliams378: Added that now

ethanmills:
How will this work with a prod account?

oskarwilliams378:
The plan is to migrate the base zone to the prod accounts, then in the prod account take the name servers of the subdomains for the non prod environments and use them to create records there. I could do that now with shared, but I thought it would be better to not have shared point to the envs, and the envs point to shared. In the end the dependency would look like
```
           |---> dev --------|
shared ----|-----------------| ---> prod
           |---> staging ----|
```

terraform/modules/data-share-service/route53.tf
```diff
@@ -0,0 +1,29 @@
...
+
+resource "aws_acm_certificate" "hosted_zone" {
+  domain_name       = aws_route53_zone.zone.name
+  validation_method = "DNS"
+}
+
+resource "aws_route53_record" "cloudfront" {
+  zone_id = aws_route53_zone.zone.zone_id
+  name    = aws_route53_zone.zone.name
+  type    = "A"
```
ethanmills, line 13: Slightly surprised this isn't a CNAME, but I've not done exactly this before

ethanmills:
As discussed we're going to leave this a little while in the hope that we can set it up in the prod account first time around and save the migration headache

oskarwilliams378:
This has been fulfilled by other PRs

PR #680, opened 2023-03-15
GPC-248: Enable Security Hub

olevett:
We probably need to do something about getting notifications of outputs, and should discuss which standards we want to follow, but the CIS standard is a generally broad reaching and sensible baseline

### DIFF
--- a/terraform/dev/main.tf
+++ b/terraform/dev/main.tf
@@ -130,8 +130,3 @@ module "consumer" {
   lev_api_url                 = "https://${module.lev_api.service_url}"
   schedule                    = "cron(0,30 9-18 ? * MON-FRI *)"
 }
-
-module "securityhub" {
-  source = "../modules/security_hub"
-  region = data.aws_region.current.name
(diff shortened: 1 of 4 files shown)


oskarwilliams378 (approved):

terraform/dev/main.tf
```diff
@@ -130,3 +130,8 @@ module "consumer" {
   lev_api_url                 = "https://${module.lev_api.service_url}"
   schedule                    = "cron(0,30 9-18 ? * MON-FRI *)"
 }
+
+module "securityhub" {
+  source = "../modules/security_hub"
+  region = data.aws_region.current.name
+}
```
ethanmills, line 137: Would this make more sense in shared?

olevett: Possibly, it depends on how we're setting up different accounts though. We'd need to enable it in prod too, as it's an account level thing.

ethanmills (approved):

PR #686, opened 2023-03-15
GPC-248: Enable another AWS standard

olevett:


ethanmills (approved):
More standards more better?

PR #692, opened 2023-03-16
GPC-NONE: Add vulnerability scanning ADR

oskarwilliams378:


ethanmills (approved):

doc/architecture/decisions/0019-vulnerability-scanning.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Approach
+
+As we now have Security Hub, we will now be using this as our central point for discovering new vulnerabilities in
+existing dependencies. All of our containers that we run, will now be gotten from our ECR, which will have continuous
+scanning. This continuous scanning will then output its findings into security hub, so we can see all the new
+developments there.
+
+For our new dependencies, we will still have Trivy scanning in our pull requests, as this will prevent the addition of
+new vulnerabilities.
```
olevett, line 30: Consequence of this is that we need to enable Security Hub in all new accounts as and when they're enabled.

We may also need to enable cross account aggregation within Security Hub

olevett: Another consequence here is that our container lifecycle policy may increase costs as it retains all images, which we should probably document.

olevett (approved):

olevett (approved):

PR #705, opened 2023-03-17
GPC-189: Fix name of iam instance profile to include environment name

CarlyG55:


terraform/modules/data-share-service/ec2.tf
```diff
@@ -96,7 +96,7 @@ data "aws_iam_policy_document" "rds_bastion_access_policy" {
 }
 
 resource "aws_iam_instance_profile" "rds_bastion_instance_profile" {
-  name = "ec2_rds"
+  name = "${var.environment}-ec2_rds"
```
oskarwilliams378, line 99:
```suggestion
  name = "${var.environment}-ec2-rds"
```

oskarwilliams378 (approved):

PR #714, opened 2023-03-20
GPC-NONE: Ban S3 public access

ethanmills:


oskarwilliams378 (approved):

terraform/shared/main.tf
```diff
@@ -106,3 +106,7 @@ module "securityhub" {
 module "ecr" {
   source = "./modules/ecr"
 }
+
+module "s3" {
```
oskarwilliams378, line 110: could we change the name of this to be like s3_policy or something, so that we don't think its just a bucket

oskarwilliams378 (changes requested):

oskarwilliams378 (approved):

PR #715, opened 2023-03-20
GPC-192: Add lifecycle policy for GDX ecr images

oskarwilliams378:


olevett (approved):

terraform/shared/modules/ecr/ecr.tf
```diff
@@ -11,6 +11,43 @@ resource "aws_ecr_repository" "gdx_data_share_poc" {
   }
 }
 
+resource "aws_ecr_lifecycle_policy" "gdx_data_share_poc" {
```
ethanmills, line 14: This doesn't seem to do quite what I'd expect. If my latest prod tagged image here is 15 days old I think this will expire it?

ethanmills (changes requested):

ethanmills (approved):

terraform/shared/modules/ecr/ecr.tf
ethanmills: Missed the untagged specifier

terraform/shared/modules/ecr/ecr.tf
```diff
@@ -19,10 +21,10 @@ resource "aws_ecr_lifecycle_policy" "gdx_data_share_poc" {
     "rules": [
         {
             "rulePriority": 1,
-            "description": "Keep last 3 tagged images",
+            "description": "Keep last 3 tagged dev images",
             "selection": {
                 "tagStatus": "tagged",
-                "tagPrefixList": ["dev", "demo", "prod"],
+                "tagPrefixList": ["demo"],
```
olevett: dev vs demo? I appreciate they're non-obvious, but I'm not sure they're that non-obvious?

oskarwilliams378: Typo by a slow oskar

ethanmills (approved):

terraform/shared/modules/ecr/ecr.tf
```diff
@@ -32,9 +34,22 @@ resource "aws_ecr_lifecycle_policy" "gdx_data_share_poc" {
...
+                "tagPrefixList": ["dev"],
+                "countType": "imageCountMoreThan",
+                "countNumber": 3
+            },
+            "action": {
+                "type": "expire"
+            }
+        },
+        {
+            "rulePriority": 3,
```
olevett, line 49: Do we need to add prod at some point when we've got it?

oskarwilliams378: Yes

PR #717, opened 2023-03-20
GPC-NONE: Remove outdated docs and add session manager installation i…

CarlyG55:
…nstruction

### DIFF
--- a/doc/connecting-to-hosted-databases.md
+++ b/doc/connecting-to-hosted-databases.md
@@ -7,6 +7,9 @@ connect to this EC2 instance via AWS Systems Manager Session Manager.
 
 ## SSM tunneling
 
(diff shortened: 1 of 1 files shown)


doc/connecting-to-hosted-databases.md
```diff
@@ -26,24 +29,3 @@ normal.
...
-```
-
-You can then connect to the database with
-
-```shell
-psql -h<host> -U<username> -d<database>
-```
-
-The host and database name can be found from the RDS section of the AWS console, while the username and password (
-entered interactively) can be found as `rds_username` and `rds_password` as above.
```
CarlyG55, line 49: Since we no longer have outbound internet access on the instance, we can't install postgresql

oskarwilliams378 (approved):

PR #731, opened 2023-03-21
GPC-NONE: Create terraform backend prod

oskarwilliams378:


terraform/backend-prod/main.tf
```diff
@@ -0,0 +1,43 @@
...
+      version = "~> 4.0"
+    }
+  }
+  #  backend "s3" {
+  #    bucket         = "gdx-data-share-tfstate"
+  #    key            = "terraform.tfstate"
+  #    region         = "eu-west-2"
+  #    dynamodb_table = "gdx-data-share-lock"
+  #    encrypt        = true
+  #  }
```
ethanmills, line 14: why comment?

ethanmills: Is that because it will fail before the bootstrap has run?

oskarwilliams378: Yep, we need to run it commented, then uncomment and run, so after it has been completed I would like a PR with it uncommented basically

PR #732, opened 2023-03-21
GPC-97: Setup initial hosted zone

oskarwilliams378:


ethanmills (approved):
This looks good. Do we have a plan for how to share the nameservers of the delegated zones (for subdomains) across to prod? Presumably there's no issue with giving prod access to the non-prod state bucket?

oskarwilliams378:
I'm not going to merge this until we have the bootstrap done and our terraform workflow in.

For delegated zones, my current plan is to give prod access to their buckets and output the names and nameservers of the hosted zones from the non prod accounts, exactly as you say.

PR #734, opened 2023-03-21
GPC-NONE: Remove unneeded s3 policies

oskarwilliams378:


ethanmills:
Trivy says no, and I'm inclined to go with it. Somehow the per bucket settings and the account settings are decoupled in the console. I assume the account settings win otherwise there's no point, but if we were ever to remove the account setting because we actually wanted a public bucket (can't think why) then having them on every bucket would be a good thing

ethanmills (approved):

ethanmills (commented):
What does the plan look like on this one? Is it a replace, or a modify?

PR #736, opened 2023-03-21
GPC-NONE: Add vpc flow logs

oskarwilliams:


terraform/modules/data-share-service/vpc.tf
```diff
@@ -9,12 +9,23 @@ module "vpc" {
   # At least two availability zones are required in order to set up a load balancer, so that the
   # infrastructure is kept consistent with other environments which use multiple availability zones.
   availability_zones = data.aws_availability_zones.available.zone_ids
-
-  tags_default = {
-    Environment = var.environment
-  }
```
oskarwilliams378, line 15: These are added by our default provider tags

olevett (approved):

PR #740, opened 2023-03-21
GPC-NONE: Allow access to cross account state

oskarwilliams378:


terraform/modules/github_iam/cross_account_access.tf
ethanmills, line 1: Is this necessary? The `ReadOnlyAccess` policy on this role has `s3:Get*` and `s3:List*` on `*`, which I think covers this?

ethanmills (approved):

terraform/modules/github_iam/cross_account_access.tf
oskarwilliams378: Am happy to give it a try

PR #751, opened 2023-03-23
GPC-251: Push prometheus-adot through ECR and code deploy

oskarwilliams378:


.github/workflows/update-prometheus.yml
```diff
@@ -0,0 +1,72 @@
...
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Install GitVersion
+        uses: gittools/actions/gitversion/setup@v0.10.2
+        with:
+          versionSpec: '5.x'
+
+      - name: Determine Version
```
olevett, line 30: Are we versioning the prometheus image?

oskarwilliams378: Will remove that here and for grafana, missed that when going through it

.github/workflows/update-prometheus.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Build and tag image
+        uses: docker/build-push-action@v4
+        with:
+          context: ./prometheus
+          push: false
+          tags: ${{ github.ref_name }}:latest
```
olevett, line 30: I don't think this is necessarily anything we need to worry about, but do we have a way to prove a heritage (e.g. this image came from X upstream version?). I think this applies to both this and the Grafana variant

olevett: I think it's plausible that just the root docker-file version + commit hash is enough here, but figured it's worth a question at least!

olevett (approved):

PR #753, opened 2023-03-23
GPC-251: Add grafana ecr

oskarwilliams378:


olevett (approved):

olevett:
I'm not _too_ worried about these having required pipelines for these, in that we expect to get similar output from AWS and we expect the vast majority of commit changes to not be touching these files. I don't think it's a massive cost to have the extra required checks, but if things slow down/we're worried about numbers of actions, we could consider it

PR #758, opened 2023-03-23
GPC-NONE: Remove trivy run on main

oskarwilliams378:


olevett:
I think the PR title is a little misleading, but otherwise 👍

PR #761, opened 2023-03-23
GPC-159: Review database user permissions

CarlyG55:


ethanmills (approved):
This looks good, thanks for sticking with it! I think next steps are  to try it out in demo. We should also think about how we can apply these in our tests.

PR #774, opened 2023-03-27
GPC-97: Setup DNS for subdomains with cloudfront

oskarwilliams:


terraform/modules/data-share-service/cloudfront.tf
```diff
@@ -89,3 +93,37 @@ resource "aws_route53_record" "cloudfront" {
...
+  type            = each.value.type
+  zone_id         = var.hosted_zone_id
+}
+
+resource "aws_acm_certificate_validation" "hosted_zone" {
+  provider = aws.us-east-1
+
+  certificate_arn         = aws_acm_certificate.hosted_zone.arn
+  validation_record_fqdns = [for record in aws_route53_record.cert_validation : record.fqdn]
+
```
oskarwilliams378, line 128:
```suggestion
```

ethanmills (approved):

PR #820, opened 2023-03-31
GPC-NONE: Update dependabot labels

oskarwilliams378:


.github/workflows/dependabot-gradle.yml
```diff
@@ -0,0 +1,33 @@
...
+      contents: write
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Setup Java
+        uses: actions/setup-java@v3
+        with:
+          distribution: corretto
+          java-version: 19
```
ethanmills, line 22: If the java update gets merged first this will need to change to 20

ethanmills (approved):

.github/workflows/dependabot-gradle.yml
```diff
@@ -0,0 +1,35 @@
...
+    if: ${{ github.event.label.name == "dependabot-gradle" && github.event.head_commit.message != "GPC-NONE: Automatically update gradle.lockfile" }}
+    name: Fix gradle lockfile for dependabot
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.PAT }}
```
olevett, line 18: We should do this for the auto-format-pr stuff too if possible

oskarwilliams378: Will do that afterwards

PR #842, opened 2023-04-03
GPC-NONE: Suppress missing xray segments

ethanmills:


ethanmills:
Not 100% sure this is correct, but it's my best interpretation of the docs: https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-java-configuration.html#xray-sdk-java-configuration-sysprops

PR #857, opened 2023-04-04
GPC-173: Use bot PAT for auto fixing code

olevett:


.github/workflows/terraform-fmt-fix.yml
```diff
@@ -16,7 +16,9 @@ jobs:
         working-directory: terraform
     steps:
       - name: Checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v3   
+        with:
+          token: ${{ secrets.PAT }}
```
ethanmills, line 21: What does this give us? I'd assume it's the autocommit action that needed the token?

oskarwilliams378: Apparently not, according to the docs for that action

ethanmills (approved):
Ok, that's just not how it works

.github/workflows/terraform-fmt-fix.yml
oskarwilliams378: https://github.com/stefanzweifel/git-auto-commit-action#commits-made-by-this-action-do-not-trigger-new-workflow-runs

olevett: As far as I can tell, it means the checkout has the ability to push the code later, through whatever action. It's a little counterintuitive though...

PR #859, opened 2023-04-04
GPC-NONE: Update java compile target

oskarwilliams378:


oskarwilliams378:
Closed as we may want to target the last lts 17

PR #873, opened 2023-04-05
GPC-252: Setup HTTPS between cloudfront and load balancers

CarlyG55:


terraform/modules/data-share-service/cloudfront.tf
```diff
@@ -12,11 +11,11 @@ resource "aws_cloudfront_distribution" "gdx_data_share_poc" {
     origin_id   = "${var.environment}-gdx-data-share-poc-lb"
 
     custom_origin_config {
+      # Required but irrelevant - we do not use HTTP to talk to LB
```
oskarwilliams378, line 14: Maybe update this comment pls

terraform/modules/data-share-service/lb.tf
```diff
@@ -15,12 +21,45 @@ resource "random_id" "http_sufix" {
...
+}
+
+resource "aws_acm_certificate_validation" "acm_lb_certificate_validation" {
+  certificate_arn         = aws_acm_certificate.acm_lb_certificate.arn
+  validation_record_fqdns = [for record in aws_route53_record.lb_cert_validation : record.fqdn]
+}
+
 #tfsec:ignore:aws-elb-http-not-used
-resource "aws_lb_listener" "listener_http" {
+resource "aws_lb_listener" "listener_https" {
```
oskarwilliams378, line 56: could you add a moved resource to say that listener_http was moved to listener_https pls (and for test_listener_http(s))

PR #890, opened 2023-04-06
GPC-263 StatusCake integration

Mjwillis:
Basic module setup done.   

Run locally and by adding to main.tf in dev was able to create an alert check.  

- Next steps to provide API token as secret.
- Parameterise other values


terraform/demo/main.tf
```diff
@@ -102,4 +103,5 @@ module "data-share-service" {
 
   hosted_zone_id   = module.route53.zone_id
   hosted_zone_name = module.route53.name
+  ping_check_url   = "https://${local.host_url}/health/ping"
```
ethanmills, line 106: Was this the one that returns 200 even when everything is broken?

Mjwillis: yes

terraform/shared/ssm.tf
```diff
@@ -0,0 +1,12 @@
+
+resource "aws_ssm_parameter" "statuscake_api_key" {
```
oskarwilliams378, line 2: Could we define this and the data for it both in the status cake module?

oskarwilliams378:
I think I would rather have the provider and the ssm resources inside of the statuscake module, make it more self reliant

terraform/shared/statuscake.tf
```diff
@@ -0,0 +1,15 @@
+data "aws_ssm_parameter" "statuscake_api_key" {
+  name = aws_ssm_parameter.statuscake_api_key.name
+}
+
+#provider "statuscake" {
```
olevett, line 5: Given this, can we make sure we document this in the new environment setup (that this will fail until the parameter is set up, but you need to run some terraform to set up SSM)

oskarwilliams378:
Could you move all the ssm parameters into the module pls

terraform/modules/statuscake/ssm.tf
```diff
@@ -0,0 +1,12 @@
+
```
oskarwilliams378, line 1:
```suggestion
```

terraform/shared/main.tf
```diff
@@ -153,3 +157,11 @@ module "iam_user_roles" {
...
   read_only_users = local.gdx_dev_team
 }
+
+#module "statuscake" {
+#  source      = "../modules/statuscake"
+#  env_url_pair =  {
+#    "dev" = "https://dev.share-life-events.service.gov.uk/health/ping"
+#    "demo" = "https://demo.share-life-events.service.gov.uk/health/ping"
+#  }
+#}
```
oskarwilliams378, line 167: This won't generate the ssm

oskarwilliams378 (approved):

PR #900, opened 2023-04-11
GPC-NONE: Remove user policy and add to group

oskarwilliams378:


terraform/modules/dwp_dev/main.tf
```diff
@@ -45,7 +44,17 @@ resource "aws_iam_policy" "dwp_user_access" {
   name   = "dwp-dev-sqs-access"
 }
 
-resource "aws_iam_user_policy_attachment" "dwp_user_access" {
+#tfsec:ignore:aws-iam-enforce-mfa
```
olevett, line 47: :( presumably because of the service-accounty nature of this the policy doesn't make much sense? I'd be keen to write a brief reason here if that's the case, rather than just ignoring it

ethanmills (commented):
Why this change?

oskarwilliams378:
It is against a lot of advice and different measures (AWS, CIS, NCSC) to have IAM policies directly attached to users

PR #902, opened 2023-04-11
Gpc 278 enable s3 server access logging

CarlyG55:


terraform/modules/bootstrap/bootstrap.tf
```diff
@@ -1,5 +1,3 @@
-# Disable S3 bucket logging as this is the only bucket, revisit when there's more underlying infra/monitoring set up
-#tfsec:ignore:aws-s3-enable-bucket-logging
 resource "aws_s3_bucket" "state_bucket" {
```
oskarwilliams378, line 1: You're going to need to add a `aws_s3_bucket_logging` resource

oskarwilliams378 (changes requested):

PR #908, opened 2023-04-11
GPC-NONE: Add IAM password policy

oskarwilliams378:


terraform/modules/policies/iam.tf
```diff
@@ -34,3 +34,14 @@ resource "aws_iam_group_policy_attachment" "enforce_mfa" {
...
+
+resource "aws_iam_account_password_policy" "password_policy" {
+  minimum_password_length        = 16
+  require_lowercase_characters   = true
+  require_numbers                = true
+  require_uppercase_characters   = true
+  require_symbols                = true
+  allow_users_to_change_password = true
+  password_reuse_prevention      = 24
+  max_password_age               = 90
```
ethanmills, line 46: Is this necessary? If this is just break glass what's the advantage in short-lived passwords?

ethanmills (approved):

terraform/modules/policies/iam.tf
oskarwilliams378: I think this is a security hardening measure that could be useful in the future, and is suggested by NCSC, so I was going to go ahead with it as it doesn't really cost us anything

PR #919, opened 2023-04-12
GPC-NONE: Suppress EC2 detailed monitoring

oskarwilliams378:


ethanmills (approved):

ci/update_securityhub_suppressions/suppressions.yml
```diff
@@ -124,3 +124,7 @@ Suppressions:
         - arn:aws:s3:::dev-lb-access-logs-logs-gdx-data-share-poc
         - arn:aws:s3:::demo-lb-access-logs-logs-gdx-data-share-poc
       notes: These are all log buckets already - we do not want to add logging to prevent infinite chain of log buckets
+  "ec2-instance-detailed-monitoring-enabled-conformance-pack-7wkwquwdu":
+    - resources:
+        - arn:aws:ec2:eu-west-2:776473272850:instance/.*
```
ethanmills, line 129: Do we want to wildcard them like this? I know it's a pain, but if it's possible to just list the two bastion hosts I think that would be preferable so this is a conscious choice we're alerted of if we add new types of instance

oskarwilliams378: So we could do that, but that would change every time the AMI image updates

PR #923, opened 2023-04-12
GPC-303: Add event notifications for S3

oskarwilliams378:


terraform/modules/grafana/variables.tf
```diff
@@ -21,3 +21,7 @@ variable "vpc_cidr" {
 variable "ecr_url" {
   type = string
 }
+
+variable "sns_topic_arn" {
```
olevett, line 25: Can we be more specific with this name? What's its' purpose as far as the Grafana setup is concerned?

olevett (approved):

PR #941, opened 2023-04-14
GPC-NONE: Use s3 module for s3 buckets

oskarwilliams378:


terraform/modules/s3/main.tf
```diff
@@ -1,9 +1,5 @@
 resource "aws_s3_bucket" "bucket" {
   bucket = "${var.environment}-${var.name}-gdx-data-share-poc"
-
-  lifecycle {
-    prevent_destroy = true
-  }
```
oskarwilliams378, line 6: This is needed to generate the plan as I need to delete the grafana cloudfront logs bucket so that it aligns properly with our naming. This will be added back in afterwards

ethanmills (approved):

PR #954, opened 2023-04-17
GPC-NONE: Fix bootstrap

oskarwilliams378:


ethanmills (approved):

terraform/modules/bootstrap/bootstrap.tf
```diff
@@ -1,21 +1,26 @@
...
 
   tiering_noncurrent_days = 10
 
   sns_arn = module.sns.topic_arn
 
+  cross_account_arns = [
+    "arn:aws:iam::255773200490:role/prod-github-oidc-deploy",
+    "arn:aws:iam::255773200490:role/github-oidc-pull-request"
+  ]
+
```
ethanmills, line 16: what are these for?

oskarwilliams378: To allow prod to access the state of non prod

PR #971, opened 2023-04-18
GPC-NONE: Add waf sql injection protection

oskarwilliams378:


terraform/modules/data-share-service/waf.tf
```diff
@@ -208,6 +208,34 @@ resource "aws_wafv2_web_acl" "load_balancer" {
...
+
+    statement {
+      sqli_match_statement {
+        text_transformation {
+          type     = "URL_DECODE"
+          priority = 1
+        }
+
+        field_to_match {
+          query_string {}
```
olevett, line 227: Does query_string match request bodies too? I feel like on our POSTs, this is worth doing

oskarwilliams378: Done

olevett (approved):

PR #994, opened 2023-04-20
GPC-NONE: Update lambda suppressions

oskarwilliams378:


ci/update_securityhub_suppressions/suppressions.yml
```diff
@@ -242,6 +242,32 @@ Suppressions:
         - arn:aws:iam::776473272850:user/dwp-dev
       notes: MFA is not enabled as this user is only a service account for access to SQS
 
+  ## Lambda
+  "lambda-dlq-check-conformance-pack-1dpd692bc":
```
olevett, line 246: Are these new to this conformance pack then, or is there an old version we should remove?

olevett (approved):

ci/update_securityhub_suppressions/suppressions.yml
olevett: Discussed offline, this is a new suppression rather than related to a change in conformance pack version

PR #997, opened 2023-04-20
GPC-NONE: Update GuardDuty suppressions

oskarwilliams378:


olevett (approved):
I take it from these that we couldn't silence them directly in Guard Duty sensibly?

PR #999, opened 2023-04-20
GPC-NONE: Add global (us-east-1) config and security hub

oskarwilliams378:


terraform/modules/security/main.tf
```diff
@@ -0,0 +1,48 @@
...
+    }
+  }
+}
+
+resource "aws_securityhub_finding_aggregator" "securityhub" {
+  depends_on   = [module.securityhub_local]
+  linking_mode = "ALL_REGIONS"
+}
+
+module "securityhub_local" {
```
olevett, line 16: "Local" here is "eu-west-2" right?

olevett (approved):

terraform/modules/security/main.tf
oskarwilliams378: With how the module works, its whatever region you set

PR #1015, opened 2023-04-21
GPC-NONE: Update documentation with data and event flows

ethanmills:


oskarwilliams378 (approved):

olevett:
The comment about "actual data flows around the system" I found a little tricky to follow, because it's right next to the callback API line, can I check if it's reference to the callback line, or to something else?

ethanmills:
> The comment about "actual data flows around the system" I found a little tricky to follow, because it's right next to the callback API line, can I check if it's reference to the callback line, or to something else?

Yes, it's a reference to the callback line @olevett

olevett:
Okay, I think "around" is the tricky word there, it feels like it's trying to say "most of the "sensitive" data flows outside of the core event platform"

PR #1045, opened 2023-04-25
GPC-358: Add initial lambda

oskarwilliams378:


terraform/modules/gro_ingestion_service/lambdas.tf
```diff
@@ -0,0 +1,69 @@
...
+resource "null_resource" "lambda-function-source-builder" {
+  provisioner "local-exec" {
+    working_dir = "${path.module}/lambdas"
+    command     = <<EOT
+      npm i
+      npm run build
+    EOT
+  }
+  triggers = {
+    lambda-function-md5 = data.archive_file.typescript-source.output_md5
```
olevett, line 30: This surprised me that it was triggered by the md5

terraform/modules/gro_ingestion_service/lambdas.tf
```diff
@@ -0,0 +1,69 @@
+locals {
+  lambda_runtime = "nodejs18.x"
+  functions = toset([
+    for file in fileset("${path.module}/lambdas/src/functions", "*.function.ts") : lower(replace(split(".", split("/", file)[length(split("/", file)) - 1])[0], "[^a-zA-Z0-9]", "-"))
+  ])
+}
+
+data "archive_file" "typescript-source" {
```
olevett, line 8:
```suggestion
data "archive_file" "typescript_source" {
```

olevett: (and others)

terraform/modules/gro_ingestion_service/lambdas.tf
```diff
@@ -0,0 +1,69 @@
...
+  type        = "zip"
+  source_dir  = "${path.module}/lambdas/src"
+  output_path = "zip/typescript-source.zip"
```
olevett, line 11: So this is a way to get a hash of the whole directory, without hashing by hand? Do we _need_ to, or can we basically just always compile it and then just diff at the end? (Not sure how the complexity here vs the build time trade off though)

terraform/modules/gro_ingestion_service/lambdas.tf
oskarwilliams378: Definitely

terraform/modules/gro_ingestion_service/lambdas.tf
```diff
@@ -0,0 +1,69 @@
...
+  source_dir  = "${path.module}/lambdas/dist"
+  output_path = "zip/lambda-function-source.zip"
+  depends_on  = [null_resource.lambda-function-source-builder]
+}
+
+resource "null_resource" "lambda-function-source-builder" {
+  provisioner "local-exec" {
+    working_dir = "${path.module}/lambdas"
+    command     = <<EOT
+      npm i
```
olevett, line 25:
```suggestion
      npm ci
```

terraform/modules/gro_ingestion_service/lambdas.tf
oskarwilliams378: Why diff after the compile rather than diff the source? diffing the source would just be quicker?

terraform/modules/gro_ingestion_service/lambdas.tf
olevett, line 1: At some point, this feels like a generic "here's a lambda from typescript" module, rather than "one exact copy of this for each TS Lambda we've got"

terraform/modules/gro_ingestion_service/lambdas.tf
oskarwilliams378: yep

terraform/modules/gro_ingestion_service/lambdas/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "tsc"
+  },
+  "devDependencies": {
+    "@types/aws-lambda": "^8.10.114",
```
olevett, line 15: I have a preference for pinning to exact versions, e.g.

```suggestion
    "@types/aws-lambda": "8.10.114",
```

there's an npmrc config setting for this I'll try and find...

terraform/modules/gro_ingestion_service/lambdas.tf
oskarwilliams378: So the purpose would be that we just add functions in the `/functions` folder and this will add all of them as lambdas

olevett (approved):

olevett:
I've left a few comments around, I don't think these need to block merging, but might be easier to deal with now rather than later.

terraform/modules/gro_ingestion_service/lambdas.tf
oskarwilliams378: Updated

terraform/modules/gro_ingestion_service/lambdas.tf
olevett: It's more complex, and I  think my question is: is compiling it that much more expensive than zipping? For something small, I suspect it'll compile very quickly (but that's a naive statement)

terraform/modules/gro_ingestion_service/lambdas/package.json
oskarwilliams378: found it

terraform/modules/gro_ingestion_service/lambdas.tf
oskarwilliams378: Compiling also involves running npm install, so I think it would be a fair bit longer imo, the zipping is incredibly quick

PR #1051, opened 2023-04-25
GPC-349: Stream DynamoDB records to GDX API

oskarwilliams378:
This adds the code for the stream lambda function, but does not add the infrastructure changes, those will be in a further PR

### DIFF
--- a/terraform/modules/gro_ingestion_service/lambdas/src/functions/publish-event.function.ts
+++ b/terraform/modules/gro_ingestion_service/lambdas/src/functions/publish-event.function.ts
@@ -51,7 +51,6 @@ const publishEvent = async (event: PublishEvent) => new Promise((resolve, reject
 
 export const handler: Handler = async (event: DynamoDBStreamEvent) => {
     const publishEvents = event.Records
-        .filter(r => r.eventName === "INSERT")
         .filter(r => r.dynamodb?.NewImage)
         .map(r => r.dynamodb?.NewImage)
         .map(convertToEvent)
--- a/terraform/modules/gro_ingestion_service/lambdas/src/functions/publish-event.function.ts
+++ b/terraform/modules/gro_ingestion_service/lambdas/src/functions/publish-event.function.ts
@@ -1,6 +1,6 @@
-import {DynamoDBRecord, DynamoDBStreamEvent, Handler} from "aws-lambda";
-import {mapToEventRecord} from "../models/EventRecord";
-import {request} from "https";
+import {DynamoDBRecord, DynamoDBStreamEvent, Handler} from "aws-lambda"
+import {mapToEventRecord} from "../models/EventRecord"
+import {request} from "https"
 
 const apiUrl = process.env.API_URL
 const options = {
@@ -11,7 +11,7 @@ const options = {
     headers: {
         "Content-Type": "application/json",
     },
-};
+}
 
 const publishRecord = async (record: DynamoDBRecord) => {
     if (record.eventName !== "INSERT" || !record.dynamodb?.NewImage) {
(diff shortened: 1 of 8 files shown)


terraform/modules/gro_ingestion_service/lambdas/src/functions/publish-event.function.ts
```diff
@@ -0,0 +1,57 @@
+import {DynamoDBRecord, DynamoDBStreamEvent, Handler} from "aws-lambda"
+import {mapToEventRecord} from "../models/EventRecord"
+import {request} from "https"
+
+const apiUrl = process.env.API_URL
+const options = {
+    hostname: apiUrl,
+    path: "/todos",
```
olevett, line 8: `todos` seems like we should probably change it!

terraform/modules/gro_ingestion_service/lambdas/src/models/EventRecord.ts
```diff
@@ -0,0 +1,45 @@
...
+    AddressLine1: string
+    AddressLine2: string
+    AddressLine3: string
+    AddressLine4: string
+    Postcode: string
+}
+
+export const mapToEventRecord = (image: { [key: string]: AttributeValue }): EventRecord => {
+    return {
+        hash: image["hash"].S ?? "",
```
olevett, line 27: `image` feels like an odd name, is there a convention I've completely missed?

terraform/modules/gro_ingestion_service/lambdas/src/functions/publish-event.function.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    const eventRecord = mapToEventRecord(record.dynamodb.NewImage)
+    const event = {
+        eventType: "DEATH_NOTIFICATION",
+        eventTime: eventRecord.EventTime,
+        id: eventRecord.hash
+    }
+
+    return new Promise((resolve, reject) => {
```
olevett, line 28: Mixing async on line 16 with promise/callbacks feels odd to me, is there a more consistent way we can write this, or is this to enable the Promise.all syntax later?

terraform/modules/gro_ingestion_service/lambdas/src/functions/publish-event.function.ts
```diff
@@ -0,0 +1,57 @@
...
+        })
+
+        req.write(event)
+        req.end()
+    })
+}
+
+export const handler: Handler = async (event: DynamoDBStreamEvent) => {
+    console.log(`Streaming ${event.Records.filter(r => r.eventName === "INSERT").length} events`)
+    await Promise.all(event.Records.map(publishRecord))
```
olevett, line 56: Do we want to `all` where we'll reject on the first rejection, or do we want to handle individual records? Also, what happens here if we do fail to call the API? I _think_ we should retry, or at least have some ability to handle it

olevett (changes requested):
This seems broadly great, couple of questions it might be easier to explore offline though

terraform/modules/gro_ingestion_service/lambdas/src/models/EventRecord.ts
oskarwilliams378: Yep, this is dynamodbs naming, they call each entry an image, so on the stream event they have a `NewImage` and an `OldImage`

terraform/modules/gro_ingestion_service/lambdas/src/functions/publish-event.function.ts
oskarwilliams378: Yep

terraform/modules/gro_ingestion_service/lambdas/src/functions/publish-event.function.ts
oskarwilliams378: Will have a further think on this

terraform/modules/gro_ingestion_service/lambdas/src/functions/publish-event.function.ts
oskarwilliams378: so returning this promise is how to make the `https` package async

terraform/modules/gro_ingestion_service/lambdas/src/functions/publish-event.function.ts
```diff
@@ -1,57 +1,72 @@
...
+
+    req.write(event)
+    req.end()
+})
 
 export const handler: Handler = async (event: DynamoDBStreamEvent) => {
-    console.log(`Streaming ${event.Records.filter(r => r.eventName === "INSERT").length} events`)
-    await Promise.all(event.Records.map(publishRecord))
+    const publishEvents = event.Records
+        .filter(r => r.eventName === "INSERT")
```
olevett, line 54: We might be able to make Lambda do these filters  for us https://docs.aws.amazon.com/lambda/latest/dg/invocation-eventfiltering.html#filtering-syntax

olevett (approved):

terraform/modules/gro_ingestion_service/lambdas/src/functions/publish-event.function.ts
```diff
@@ -0,0 +1,105 @@
...
+import { request, RequestOptions } from "https"
+import { PublishEvent } from "../models/PublishEvent"
+import { AttributeValue } from "aws-lambda/trigger/dynamodb-stream"
+
+const apiUrl = process.env.API_URL
+const authUrl = process.env.AUTH_URL
+const clientId = process.env.CLIENT_ID
+const clientSecret = process.env.CLIENT_SECRET
+
+const makeRequest = async (options: RequestOptions, requestData: string): Promise<object> => new Promise((resolve, reject) => {
```
olevett, line 12: I'd be interested to have _some_ form of testing around these, as a bit of a framework for how we test more complicated things

olevett (approved):

PR #1059, opened 2023-04-26
GPC-349: Add infrastructure setup in

oskarwilliams378:


ci/update_securityhub_suppressions/suppressions.yml
```diff
@@ -302,6 +304,8 @@ Suppressions:
         - arn:aws:lambda:eu-west-2:776473272850:function:dev-get-event
         - arn:aws:lambda:eu-west-2:776473272850:function:dev-get-events
         - arn:aws:lambda:eu-west-2:776473272850:function:dev-delete-event
+        - arn:aws:lambda:eu-west-2:776473272850:function:dev-gro-ingestion-lambda-function-publish-event
+        - arn:aws:lambda:eu-west-2:776473272850:function:demo-gro-ingestion-lambda-function-publish-event
```
ethanmills, line 308: I don't think this note applies to these functions?

terraform/modules/gro_ingestion_service/publish_event.tf
```diff
@@ -0,0 +1,39 @@
...
+  handler = "index.handler"
+  runtime = local.lambda_runtime
+  role    = aws_iam_role.publish_event_lambda.arn
+  timeout = 10
+
+  source_code_hash = data.archive_file.lambda_function_source.output_sha
+
+  environment {
+    variables = {
+      "FUNCTION_NAME" = "publishEvent"
```
ethanmills, line 29: Is this used?

oskarwilliams378: This is what lets the code get triggered via the index

ci/update_securityhub_suppressions/suppressions.yml
oskarwilliams378: I've removed test, but it does. The lambdas are proxies that only access our system over the public internet

terraform/demo/main.tf
```diff
@@ -109,4 +109,9 @@ module "gro_ingestion_service" {
 
   environment = local.env
   region      = data.aws_region.current.name
+
+  gdx_url                 = module.data_share_service.gdx_url
+  auth_url                = module.data_share_service.token_auth_url
+  publisher_client_id     = module.data_share_service.gro_ingestion_client_id
+  publisher_client_secret = module.data_share_service.gro_ingestion_client_secret
```
olevett, line 116: Is this the raw credential passing around/will it end up tfstate? If so, is there any sensible way we can use SSM/inject the secret without knowing it as much?

olevett (approved):

terraform/demo/main.tf
oskarwilliams378: It will end up in tfstate, I can mark it as sensitive, but this is the same as the current setup for our consumer.

PR #1064, opened 2023-04-27
GPC-350 Lambda to retrieve enriched data from DynamoDB

Mjwillis:


terraform/modules/gro_ingestion_service/lambdas/package.json
```diff
@@ -13,6 +13,7 @@
   },
   "devDependencies": {
     "@types/aws-lambda": "8.10.114",
+    "aws-sdk": "2.1366.0",
```
oskarwilliams378, line 16: can we move all the dependencies into normal ones and not dev ones

terraform/modules/gro_ingestion_service/lambdas/src/functions/enrichEvent.function.ts
```diff
@@ -0,0 +1,34 @@
...
+        return {
+            statusCode: 404
+        }
+    }
+
+    return {
+        statusCode: 200,
+        event: mapToEventRecord(result.Item)
+    }
+
```
oskarwilliams378, line 33: extra line

terraform/modules/gro_ingestion_service/lambdas/src/functions/enrichEvent.function.ts
```diff
@@ -0,0 +1,34 @@
...
+    const result = await dynamo.get(params).promise()
+
+    if (!result.Item) {
+        console.log(`Record with hash ${event.id} not found`)
+        return {
+            statusCode: 404
+        }
+    }
+
```
oskarwilliams378, line 28: could we add a log that we have successfully enriched this event

also could we add logParams like we have in https://github.com/alphagov/gdx-data-share-poc/pull/1065/files#diff-2ba6df16e5593d034873a60f8eceb033076b2b9bdaad8af5753ed63c0e5a0cc1 which have the EventTime, RegistrationId, and hash

oskarwilliams378 (changes requested):

oskarwilliams378 (approved):

PR #1072, opened 2023-04-28
GPC-347 step function xml to dynamodb

CarlyG55:


terraform/modules/gro_ingestion_service/lambdas/src/functions/mapXml.function.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const pushRecord = async (record: PutItemCommandInput) => {
+    const command = new PutItemCommand(record)
+    await client.send(command)
+}
+
+const generateRecord = (deathRegistration: GroDeathRegistration): PutItemCommandInput => {
+    const eventHash = hash(deathRegistration)
+    return {
+        Item: mapToDynamoDbItem(deathRegistration, eventHash),
```
oskarwilliams378, line 19: Could we do the hashing in the mapping?

terraform/modules/gro_ingestion_service/lambdas/src/models/GroDeathRegistration.ts
```diff
@@ -0,0 +1,68 @@
...
+    DeceasedAddress: {
+        Flat: string,
+        Building: string,
+        Line: string[],
+        Postcode: string,
+    },
+}
+
+
+export const mapToDynamoDbItem = (deathRecord: GroDeathRegistration, hash: string): { [key: string]: AttributeValue } => {
```
oskarwilliams378, line 48: I would rather we map to EventRecord, then we can run `marshal(eventRecord)` to get the dynamoDbItem. marshal is in `@aws-sdk/util-dynamodb`

terraform/modules/gro_ingestion_service/lambdas/src/models/EventRecord.ts
```diff
@@ -43,3 +44,28 @@ export const mapToEventRecord = (image: { [key: string]: AttributeValue }): Even
         Postcode: image["Postcode"].S ?? "",
     }
 }
+
```
oskarwilliams378, line 47: for the one above here we can do `unmarshal(image) as EventRecord` as using `@aws-sdk/util-dynamodb`

terraform/modules/gro_ingestion_service/lambdas/src/functions/splitXml.function.ts
```diff
@@ -0,0 +1,28 @@
...
+        Bucket: event.Records[0].s3.bucket.name,
+        Key: event.Records[0].s3.object.key
+    })
+    const groFileResponse = await client.send(getGroCommand)
+
+    return groFileResponse.Body?.transformToString()
+}
+
+const parseGroXml = (groXml: string): GroXmlFile => {
+    const parser = new XMLParser()
```
oskarwilliams378, line 19: Can we create this and the S3 client outside of the functions

terraform/modules/gro_ingestion_service/lambdas/src/functions/splitXml.function.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    return parser.parse(groXml)
+}
+
+export const handler: Handler = async (event: S3Event) => {
+    const groFile = await getGroFile(event)
```
oskarwilliams378, line 25: if this is null or empty string, we should probably error

terraform/modules/gro_ingestion_service/lambdas/src/models/GroXmlFile.ts
```diff
@@ -0,0 +1,6 @@
+import { GroDeathRegistration } from "./GroDeathRegistration"
+
+export interface GroXmlFile {
```
oskarwilliams378, line 3: This isn't the XML file, this is the JSON object of it, could you pls try a rename (naming is hard :( )

oskarwilliams378 (changes requested):

terraform/modules/gro_ingestion_service/lambdas/src/models/GroXmlFile.ts
CarlyG55: It's not needed anymore whoop

terraform/modules/gro_ingestion_service/lambdas/src/functions/splitXml.function.ts
```diff
@@ -0,0 +1,35 @@
...
+            error: `File with key ${event.Records[0].s3.object.key} not found`,
+        }
+        console.error("Failed to insert records into DynamoDB", logParams)
+        return {
+            statusCode: 404,
+        }
+    }
+
+    const groJson = parser.parse(groXml)
+    return { deathRegistrations: groJson.deathRegistrationGroup }
```
CarlyG55, line 34: Not sure if we now want to add a status code here?

oskarwilliams378: This is all happening inside lambdas from S3 to DynamoDB, I don't think we need a status code

PR #1073, opened 2023-04-28
GPC-351 Delete Event Function

Mjwillis:


terraform/modules/gro_ingestion_service/delete_event.tf
```diff
@@ -0,0 +1,76 @@
...
+      "kms:ReEncryptTo"
+    ]
+    effect = "Allow"
+    resources = [
+      aws_kms_key.gro_ingestion.arn
+    ]
+  }
+
+  statement {
+    sid = "DynamoDBGetItemAccess"
```
oskarwilliams378, line 24: this may be wrong

terraform/modules/gro_ingestion_service/lambdas/test/functions/dbItem.ts
```diff
@@ -0,0 +1,53 @@
+export const dbItem = {
```
oskarwilliams378, line 1: Should we add this to a mocks or consts folder?

terraform/modules/gro_ingestion_service/lambdas/test/functions/deleteEvent.function.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+
+import { DynamoDBClient, dynamoDbSendFn } from "../__mocks__/@aws-sdk/client-dynamodb"
+
+const db = new DynamoDBClient()
+
+describe("Unit test for delete event handler", function () {
+    test("verifies successful response", async () => {
+        dynamoDbSendFn.mockReturnValueOnce(Promise.resolve({ Item: dbItem }))
+
+        const event: EventRequest = {
```
oskarwilliams378, line 16: we're using this a few times, should this also be a const?

terraform/modules/gro_ingestion_service/lambdas/src/functions/deleteEvent.function.ts
```diff
@@ -0,0 +1,54 @@
...
+    try {
+        const command = new DeleteItemCommand(params)
+        const result = await dynamo.send(command)
+
+        let logParams = {
+            hash: event.id,
+            RegistrationId: "",
+            EventTime: "",
+        }
+        if (result.Attributes) {
```
oskarwilliams378, line 30: How would we get to the case where there are no Attributes here?

terraform/modules/gro_ingestion_service/lambdas/src/functions/deleteEvent.function.ts
```diff
@@ -0,0 +1,54 @@
...
+            logParams = {
+                ...logParams,
+                RegistrationId: eventRecord?.RegistrationId,
+                EventTime: eventRecord?.EventTime,
+            }
+        }
+        console.log("Successfully deleted event", logParams)
+        return {
+            id: event.id,
+            statusCode: 204,
```
oskarwilliams378, line 41: as we are returning an ID would this be a 200 rather than 204?

oskarwilliams378 (changes requested):

terraform/modules/gro_ingestion_service/lambdas/src/functions/deleteEvent.function.ts
Mjwillis: It was just a edge case where the attributes where not returned but probs not possible

terraform/modules/gro_ingestion_service/lambdas/test/functions/dbItem.ts
Mjwillis: yes

terraform/modules/gro_ingestion_service/lambdas/src/functions/deleteEvent.function.ts
Mjwillis: yes

terraform/modules/gro_ingestion_service/lambdas/test/functions/deleteEvent.function.spec.ts
Mjwillis: Actually its different

oskarwilliams378: Here and in enrichEvent it looks to be the below, whats the difference?
```
        const event: EventRequest = {
            id: "hash1",
        }
```

terraform/modules/gro_ingestion_service/lambdas/test/const/dbItem.ts
```diff
@@ -0,0 +1,53 @@
...
+        S: "Forename",
+    },
+    Surname: {
+        S: "Surname",
+    },
+    Sex: {
+        S: "Male",
+    },
+    MaidenSurname: {
+        S: null,
```
oskarwilliams378, line 24: I think these should be empty string not null, as we will want to make sure the record is returned with that and is stored with that

oskarwilliams378 (changes requested):
A couple comments left :)

PR #1082, opened 2023-05-02
GPC-273: Notifications for admin actions

ethanmills:


ethanmills:
Shout early if you're unhappy with the approach before I apply it everywhere

src/main/kotlin/uk/gov/gdx/datashare/services/AdminService.kt
```diff
@@ -12,12 +12,14 @@ class AdminService(
...
   private val acquirersService: AcquirersService,
   private val suppliersService: SuppliersService,
+  private val adminActionAlertsService: AdminActionAlertsService,
 ) {
   fun createCognitoClient(cognitoClientRequest: CognitoClientRequest) =
     cognitoService.createUserPoolClient(cognitoClientRequest)
 
   @Transactional
   fun createAcquirer(createAcquirerRequest: CreateAcquirerRequest): CognitoClientResponse? {
+    adminActionAlertsService.noticeAction(AdminAction("Create Acquirer", createAcquirerRequest))
```
oskarwilliams378, line 22: Would we want to alert on success or attempt? I fear that if this fails we would still get the alert saying that it had been created.

src/main/kotlin/uk/gov/gdx/datashare/services/AdminActionAlertsService.kt
```diff
@@ -0,0 +1,43 @@
...
+import software.amazon.awssdk.services.sns.model.PublishRequest
+
+@Service
+class AdminActionAlertsService(
+  @Value("\${admin-action-alert-sns-topic-arn:#{null}}") private val topicArn: String?,
+  private val objectMapper: ObjectMapper,
+) {
+  private val snsClient by lazy { SnsClient.create() }
+
+  fun noticeAction(action: AdminAction) {
```
oskarwilliams378, line 18: Is this the right method name?

oskarwilliams378 (commented):
2 comments, but I like the overall approach

oskarwilliams378 (approved):

PR #1097, opened 2023-05-04
GPC-NONE: Fix lambda for building

oskarwilliams378:


terraform/modules/gro_ingestion_service/lambdas/src/index.ts
```diff
@@ -1,28 +1,29 @@
...
+        deleteEvent,
+        deleteXml,
+        enrichEvent,
+        mapXml,
+        publishEvent,
+        splitXml,
     ]
 
-    const lambdaFunction = functions.find(f => f.name === `${functionName}Handler`)
+    const lambdaFunction = functions.find(f => f.name === `${functionName}`)
```
oskarwilliams378, line 16:
```suggestion
    const lambdaFunction = functions.find(f => f.name === functionName)
```

olevett (approved):

terraform/modules/gro_ingestion_service/lambdas/src/index.ts
```diff
@@ -1,28 +1,22 @@
...
     // eslint-disable-next-line @typescript-eslint/no-var-requires
-    const functions = [
-        deleteEventHandler,
-        deleteXmlHandler,
-        enrichEventHandler,
-        mapXmlHandler,
-        publishEventHandler,
-        splitXmlHandler,
-    ]
+    const functions: LambdaFunction[] = [deleteEvent, deleteXml, enrichEvent, mapXml, publishEvent, splitXml]
```
olevett, line 14: I'd be tempted to move this into an export in a new `functions/index.ts`,  but that's a very week opinion

PR #1100, opened 2023-05-04
GPC-NONE: Fix publish event lambda

oskarwilliams378:


terraform/modules/gro_ingestion_service/lambdas/src/functions/publishEvent.ts
```diff
@@ -39,7 +43,7 @@ const makeRequest = async (url: string, options: RequestOptions, requestData: st
     })
 
 const getAccessToken = async () => {
-    const authRequest = JSON.stringify({
```
ethanmills, line 42: why do we need to do this?

oskarwilliams378: The auth request has to be a url encoded form, which means it needs to be querystring and not JSON

ethanmills (approved):

PR #1104, opened 2023-05-05
GPC-NONE: Add gro death notification

oskarwilliams378:


src/main/kotlin/uk/gov/gdx/datashare/enums/GroSex.kt
```diff
@@ -0,0 +1,8 @@
+package uk.gov.gdx.datashare.enums
+
+import com.fasterxml.jackson.annotation.JsonValue
+
+enum class GroSex(@JsonValue val jsonName: String) {
```
ethanmills, line 5: Do we want to pass this on directly, or transform it?

src/main/kotlin/uk/gov/gdx/datashare/models/GroDeathNotificationDetails.kt
```diff
@@ -0,0 +1,175 @@
...
+import java.time.LocalDate
+
+@Schema(description = "GRO Death notification")
+@JsonSerialize(using = GroDeathNotificationDetailsSerializer::class)
+data class GroDeathNotificationDetails(
+  val enrichmentFields: List<EnrichmentField>,
+
+  @Schema(description = "Registration ID of death record", required = true, example = "123456789")
+  val registrationId: String? = null,
+
```
ethanmills, line 21: What does it mean to be nullable but required?

ethanmills (approved):

PR #1124, opened 2023-05-10
GPC-NONE: Add security contact phone number ssm param

oskarwilliams:


terraform/modules/policies/alternate_accounts.tf
```diff
@@ -0,0 +1,25 @@
...
+  lifecycle {
+    ignore_changes = [
+      value
+    ]
+  }
+}
+
+data "aws_ssm_parameter" "security_contact_number" {
+  name = "DWPPrincipal"
+}
```
ethanmills, line 25: This complained when I did them both at the same time. Can you add a `depends_on` and see if that works please?

PR #1125, opened 2023-05-10
GPC-NONE: Add off-boarding ADR

oskarwilliams378:


doc/architecture/decisions/0025-client-off-boarding.md
```diff
@@ -0,0 +1,38 @@
...
+2. Delete acquirer subscription
+3. Delete supplier
+4. Delete supplier subscription
+
+Deleting the subscriptions is in case a supplier or acquirer would like to maintain another subscription, but remove
+this one. In these cases, we will:
+
+1. Mark the subscription as deleted, so that no new events get sent to/from it
+2. For acquirer subscriptions, mark all non-consumed acquirer events as consumed
+3. If there is a queue connected to the subscription, tag the queue as DeletedAt (with the date and time)
```
ethanmills, line 31: Why tag and not delete the queue?

oskarwilliams378: My reasoning was that we would follow the same process as with the cognito clients, so that we don't lose any data.

I was thinking that if we tag them as deleted, then if we need to recover them due to an error we do not need to reissue keys

doc/architecture/decisions/0025-client-off-boarding.md
```diff
@@ -0,0 +1,46 @@
...
+2. Delete acquirer subscription
+3. Delete supplier
+4. Delete supplier subscription
+
+Deleting the subscriptions is in case a supplier or acquirer would like to maintain another subscription, but remove
+this one. In these cases, we will:
+
+1. Mark the subscription as deleted, so that no new events get sent to/from it
+2. For acquirer subscriptions, mark all non-consumed acquirer events as consumed
+3. If there is a queue connected to the subscription, tag the queue as whenDeleted (with the date and time)
```
ethanmills, line 39: Should we also purge the queue?

Mjwillis (approved):
LGTM

PR #1127, opened 2023-05-10
GPC-368: Add when deleted to clients and subscriptions

oskarwilliams378:


src/main/kotlin/uk/gov/gdx/datashare/repositories/Acquirer.kt
```diff
@@ -24,6 +24,8 @@ data class Acquirer(
   val name: String,
   @Schema(description = "Indicates when the Acquirer was created", required = true, example = "2023-01-04T12:30:00")
   val whenCreated: LocalDateTime = LocalDateTime.now(),
+  @Schema(description = "Indicates when the Acquirer was deleted", required = false, example = "2023-01-04T12:30:00")
+  val whenDeleted: LocalDateTime? = null,
```
ethanmills, line 28: We've currently got a mix of whenX and xAt. Which one are we going for? I thought it was xAt?

oskarwilliams378: As it is whenCreated, I thought I should add whenDeleted. Can change that if you think so?

ethanmills (approved):

PR #1130, opened 2023-05-10
GPC-NONE: ADR for persisting acquirer events early

ethanmills:


doc/architecture/img/event-flow.drawio.svg
oskarwilliams378, line 1: Do you want to update the persisting of events in the supplier event processor here?

doc/architecture/decisions/0025-internal-event-flow.md
```diff
@@ -0,0 +1,35 @@
...
+delete it, consistent with our principals of minimising data retention.
+
+Currently, it is not possible to establish if an event has been fully consumed or if further acquirer events will be created
+just from looking at the database.
+
+## Decision
+We have decided to slightly modify the existing fan-out of events to ensure all acquirer events are persisted to the database
+at the same time as the supplier event. This means that it is always possible to determine if the source event data should
+be deleted just by looking at the state of the acquirer events in the database. Once all acquirer events have been consumed,
+or an agreed length of time has passed, the data can be deleted.
```
Mjwillis, line 23: When should we check for "all consumed" records?  If we remove once time has passed this means we need to schedule a job for this.  Therefore does it make sense to just check for both time passing and consumed events on an periodic basis rather than on events

Mjwillis (approved):
Some comments

PR #1138, opened 2023-05-11
GPC-368: Delete suppliers

oskarwilliams378:


src/main/kotlin/uk/gov/gdx/datashare/repositories/Supplier.kt
```diff
@@ -17,9 +16,9 @@ import java.util.*
 @Schema(description = "Supplier")
 data class Supplier(
   @Id
-  @Column("id")
+  @JvmField
```
olevett, line 19: I'm not sure I fully understand this change

src/main/kotlin/uk/gov/gdx/datashare/services/SuppliersService.kt
```diff
@@ -124,4 +124,27 @@ class SuppliersService(
...
+        object {
+          val supplierId = id
+          val whenDeleted = now
+        },
+      ),
+    )
+    val supplier = supplierRepository.save(
+      supplierRepository.findByIdOrNull(id)?.copy(
+        whenDeleted = now,
+      ) ?: throw SupplierSubscriptionNotFoundException("Supplier $id not found"),
```
olevett, line 144: I'm not convinced this should be an error - if you call delete twice should it return OK because the entity has been deleted?

src/main/kotlin/uk/gov/gdx/datashare/services/SuppliersService.kt
```diff
@@ -124,4 +124,27 @@ class SuppliersService(
...
+  fun deleteSupplier(
+    id: UUID,
+  ): Supplier {
+    val now = dateTimeHandler.now()
+    adminActionAlertsService.noticeAction(
+      AdminAction(
+        "Delete supplier",
+        object {
+          val supplierId = id
+          val whenDeleted = now
```
ethanmills, line 137: This probably isn't necessary

src/main/kotlin/uk/gov/gdx/datashare/services/SuppliersService.kt
```diff
@@ -124,4 +124,27 @@ class SuppliersService(
...
+      ),
+    )
+    val supplier = supplierRepository.save(
+      supplierRepository.findByIdOrNull(id)?.copy(
+        whenDeleted = now,
+      ) ?: throw SupplierSubscriptionNotFoundException("Supplier $id not found"),
+    )
+    val subscriptions = supplierSubscriptionRepository.findAllBySupplierId(id)
+    subscriptions.forEach { deleteSupplierSubscription(it.id) }
+    return supplier
```
olevett, line 148: Why do we need to return the supplier?

olevett (approved):

src/main/kotlin/uk/gov/gdx/datashare/services/SuppliersService.kt
```diff
@@ -124,4 +124,27 @@ class SuppliersService(
...
+      AdminAction(
+        "Delete supplier",
+        object {
+          val supplierId = id
+          val whenDeleted = now
+        },
+      ),
+    )
+    val supplier = supplierRepository.save(
+      supplierRepository.findByIdOrNull(id)?.copy(
```
ethanmills, line 142: Should we just make `whenDeleted` mutable?

ethanmills: If we're super keen on this being immutable, how would you feel about a `UPDATE ... RETURNING` repository method?

PR #1141, opened 2023-05-11
GPC-352 emit callback/event on consumption of events to delete

Mjwillis:


src/main/kotlin/uk/gov/gdx/datashare/services/EventConsumedCheckingService.kt
```diff
@@ -0,0 +1,50 @@
...
+  val id: String,
+)
+
+data class DeleteEventResponse(
+  val id: String,
+  val statusCode: Int,
+)
+
+interface DeleteEventService {
+  @LambdaFunction(functionName = "dev-gro-ingestion-lambda-function-delete-event")
```
Mjwillis, line 48: This isn't the real way obs - as need to setup dev/demo etc

src/main/kotlin/uk/gov/gdx/datashare/services/EventConsumedCheckingService.kt
```diff
@@ -0,0 +1,50 @@
...
+import uk.gov.gdx.datashare.repositories.SupplierEventRepository
+import java.util.*
+
+@Service
+class EventConsumedCheckingService(
+  private val acquirerEventRepository: AcquirerEventRepository,
+  private val supplierEventRepository: SupplierEventRepository,
+) {
+
+  @Transactional
```
Mjwillis, line 18: Will need to sort out transactional stuff too

build.gradle.kts
```diff
@@ -73,6 +73,8 @@ dependencies {
   implementation("net.javacrumbs.shedlock:shedlock-spring:5.2.0")
   implementation("net.javacrumbs.shedlock:shedlock-provider-jdbc-template:5.2.0")
 
+  implementation("com.amazonaws:aws-java-sdk-lambda:1.12.465")
```
oskarwilliams378, line 76: Could we add this to the other Aws implementations up higher

src/main/kotlin/uk/gov/gdx/datashare/repositories/AcquirerEventRepository.kt
```diff
@@ -62,4 +62,12 @@ interface AcquirerEventRepository : CrudRepository<AcquirerEvent, UUID> {
...
 
   fun countByDeletedAtIsNull(): Int
+
+  @Query(
+    "SELECT ae.supplier_event_id  " +
+      "FROM acquirer_event ae JOIN supplier_event se on ae.supplier_event_id = se.id and se.event_consumed = false " +
+      "GROUP BY ae.supplier_event_id " +
+      "HAVING SUM( CASE when ae.deleted_at is null then 1 else 0 END ) = COUNT(*)",
+  )
+  fun findAllByDeletedEventsForAllAcquirers(): List<UUID>
```
oskarwilliams378: I think this is only going to return the supplier event IDs, so should this be in the acquirer event repo? 

Also for the logic, could we just make the sum be add 1 for each event not deleted, and then we check for it equaling 0 not Count(*)

src/main/kotlin/uk/gov/gdx/datashare/services/ScheduledConsumedEventCheckerService.kt
```diff
@@ -0,0 +1,20 @@
...
+import org.springframework.stereotype.Service
+import java.util.concurrent.TimeUnit
+
+@Service
+class ScheduledConsumedEventCheckerService(
+  private val eventConsumedCheckingService: EventConsumedCheckingService,
+) {
+
+  @Scheduled(fixedRate = 1, timeUnit = TimeUnit.HOURS)
+  @SchedulerLock(name = "checkConsumedEvents", lockAtMostFor = "3m", lockAtLeastFor = "3m")
```
oskarwilliams378, line 15: Are these the lock time periods we want? I'm unsure of how to calculate them

src/main/resources/db/migration/V9__consumed_event_marker.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE supplier_event
+    ADD COLUMN event_consumed BOOLEAN NOT NULL DEFAULT FALSE;
```
oskarwilliams378, line 2: Do we want to add migration logic to mark all the ones that should be consumed as consumed?

src/main/kotlin/uk/gov/gdx/datashare/repositories/AcquirerEventRepository.kt
```diff
@@ -62,4 +62,12 @@ interface AcquirerEventRepository : CrudRepository<AcquirerEvent, UUID> {
...
 
   fun countByDeletedAtIsNull(): Int
+
+  @Query(
+    "SELECT ae.supplier_event_id  " +
+      "FROM acquirer_event ae JOIN supplier_event se on ae.supplier_event_id = se.id and se.event_consumed = false " +
+      "GROUP BY ae.supplier_event_id " +
+      "HAVING SUM( CASE when ae.deleted_at is null then 0 else 1 END ) = COUNT(*)",
+  )
+  fun findAllByDeletedEventsForAllAcquirers(): List<UUID>
```
oskarwilliams378, line 72: As this is returning supplier event IDs, should it be in the supplier event repo?

Also for the sum logic, could we have 1 when deleted at is null, and then instead we check that the value is 0 not count(*)?

build.gradle.kts
```diff
@@ -84,7 +86,6 @@ dependencies {
   testImplementation(platform("org.testcontainers:testcontainers-bom:1.18.1"))
   testImplementation("org.testcontainers:localstack")
   testImplementation("org.testcontainers:postgresql")
-  testImplementation("com.amazonaws:aws-java-sdk-core:1.12.467") // required for TestContainers https://github.com/testcontainers/testcontainers-java/issues/1442#issuecomment-694342883
```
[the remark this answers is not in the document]
Mjwillis: Think its bundled with lambda library

src/main/kotlin/uk/gov/gdx/datashare/repositories/AcquirerEventRepository.kt
Mjwillis: yes better

src/main/resources/db/migration/V9__consumed_event_marker.sql
Mjwillis: yes good point

src/main/kotlin/uk/gov/gdx/datashare/services/ScheduledConsumedEventCheckerService.kt
Mjwillis: Not sure what we want yet, probs need to think about how long to keep lock open for and how often to run

src/main/kotlin/uk/gov/gdx/datashare/repositories/AcquirerEventRepository.kt
Mjwillis: yes

build.gradle.kts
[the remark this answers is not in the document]
Mjwillis: I've swapped to @CarlyG55 library

src/main/kotlin/uk/gov/gdx/datashare/services/GroApiService.kt
```diff
@@ -4,18 +4,34 @@ import com.fasterxml.jackson.databind.ObjectMapper
...
   private val objectMapper: ObjectMapper,
-  @Value("\${enrich-event-lambda-function-name:#{null}}") val functionName: String?,
+  private val supplierEventRepository: SupplierEventRepository,
+  @Value("\${delete.event.lambda.function.name:#{null}}") val deleteFunctionName: String?,
+  @Value("\${enrich-event-lambda-function-name:#{null}}") val enrichFunctionName: String?,
 ) {
+  fun deleteConsumedGroSupplierEvents() {
+    val now = dateTimeHandler.now()
+    supplierEventRepository.findGroDeathEventsForDeletion()
+      .forEach {
```
ethanmills, line 27: discussed breaking this down in the scheduler so that the transaction wraps each item rather than the list

src/main/kotlin/uk/gov/gdx/datashare/repositories/SupplierEventRepository.kt
```diff
@@ -1,8 +1,20 @@
...
 import java.util.*
 
 @Repository
-interface SupplierEventRepository : CrudRepository<SupplierEvent, UUID>
+interface SupplierEventRepository : CrudRepository<SupplierEvent, UUID> {
+  @Query(
+    "SELECT se FROM supplier_event se " +
+      "JOIN supplier_subscription ss ON ss.id = se.supplier_subscription_id " +
+      "LEFT JOIN acquirer_event ae ON ae.supplier_event_id = se.id " +
+      "WHERE se.deleted_at IS NULL " +
```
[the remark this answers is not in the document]
oskarwilliams378: We do not at the moment, I was expecting retention like this would be enforced in the GRO dynamo db and not here

src/main/kotlin/uk/gov/gdx/datashare/repositories/SupplierEventRepository.kt
```diff
@@ -1,8 +1,20 @@
...
+      "AND ss.event_type = 'GRO_DEATH_NOTIFICATION'" +
+      "GROUP BY se.id " +
+      "HAVING SUM( CASE WHEN ae.deleted_at IS NULL THEN 1 ELSE 0 END ) = 0",
```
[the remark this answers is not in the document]
oskarwilliams378: Happy to go your way

src/main/kotlin/uk/gov/gdx/datashare/services/GroApiService.kt
oskarwilliams378: Working on it

src/main/kotlin/uk/gov/gdx/datashare/helpers/TimeLimitedRepeater.kt
```diff
@@ -0,0 +1,19 @@
...
+      val startTime = LocalDateTime.now()
+      items.forEach {
+        repeatableFunction(it)
+        if (LocalDateTime.now() > startTime.plusSeconds(timeLimitInSeconds ?: defaultTimeLimit)) {
+          return
+        }
+      }
+    }
+  }
+}
```
ethanmills, line 19: Is there a reason we've got this in a class?
```suggestion
fun <T> repeatForLimitedTime(items: Iterable<T>, action: (item: T) -> Unit, timeLimitInSeconds: Long = 45) {
  val startTime = LocalDateTime.now()
  items.forEach {
    action(it)
    if (LocalDateTime.now() > startTime.plusSeconds(timeLimitInSeconds)) {
      return
    }
  }
} 
```

src/main/kotlin/uk/gov/gdx/datashare/repositories/SupplierEventRepository.kt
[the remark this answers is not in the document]
ethanmills: I think we probably ought to expire the acquirer events when we no longer have the data

src/main/kotlin/uk/gov/gdx/datashare/services/ScheduledJobService.kt
```diff
@@ -23,4 +28,14 @@ class ScheduledJobService(
     val unconsumedEventsCount = acquirerEventRepository.countByDeletedAtIsNull()
     gauge!!.set(unconsumedEventsCount)
   }
+
+  @Scheduled(fixedRate = 1, timeUnit = TimeUnit.MINUTES)
+  @SchedulerLock(name = "deleteConsumedGroSupplierEvents", lockAtMostFor = "45s", lockAtLeastFor = "15s")
```
ethanmills, line 33: The `lockAtMostFor` should be a bit more than the time limit. It might take ~ 1 second to build the list of events and finish up the last one to be processed. I'd set this to 60s

src/test/kotlin/uk/gov/gdx/datashare/services/ScheduledJobServiceTest.kt
```diff
@@ -0,0 +1,64 @@
...
+    val supplierEvents = listOf(supplierEvent1, supplierEvent2)
+    every { supplierEventRepository.findGroDeathEventsForDeletion() } returns supplierEvents
+    every { TimeLimitedRepeater.repeat(supplierEvents, groApiService::deleteConsumedGroSupplierEvent) } just runs
+
+    underTest.deleteConsumedGroSupplierEvents()
+
+    verify(exactly = 1) {
+      TimeLimitedRepeater.repeat(supplierEvents, groApiService::deleteConsumedGroSupplierEvent)
+    }
+  }
```
ethanmills, line 62: This feels a bit like testing the implementation. I think I'd assert that it calls `deleteConsumedGroSupplierEvent` with each event

ethanmills (approved):

PR #1161, opened 2023-05-16
GPC-270: create acquirer events atomically

ethanmills:


src/main/kotlin/uk/gov/gdx/datashare/services/AcquirerEventProcessor.kt
```diff
@@ -23,32 +25,29 @@ class AcquirerEventProcessor(
...
+
     val acquirerSubscription = acquirerSubscriptionRepository.findByIdOrNull(acquirerEvent.acquirerSubscriptionId)
       ?: throw AcquirerSubscriptionNotFoundException(acquirerEvent.acquirerSubscriptionId)
 
-    if (acquirerSubscription.queueName != null) {
-      persistEventAndPushToQueue(acquirerEvent, acquirerSubscription)
-    } else {
-      persistEvent(acquirerEvent)
+    if (acquirerSubscription.queueName == null) {
+      return
```
olevett, line 38: Love an early return, though curious as to why this can be null/could we just filter it out in the acquirerSubscription call?

olevett (approved):

PR #1182, opened 2023-05-18
GPC-NONE: Make supplier event idempotent

oskarwilliams:


olevett:
Is this going to be okay on dev? Are our current IDs unique?

oskarwilliams378:
Will test this now, I think yes but better safe

olevett:
As a note, we don't want to merge this right now as it'll cause issues in a couple of places
- our automated deployment testing
- we have some duplicated data so the migration may fail

PR #1189, opened 2023-05-19
GPC-383: Code to create acquirer queues

ethanmills:
The first of several steps to manage acquirer queues from the application. This allows standard and FIFO queues to be created from the application.

This is me finishing up some work I started a while ago. Remaining steps are to modify the admin endpoints to allow this to be set up from the API, and document

### DIFF
--- a/src/test/kotlin/uk/gov/gdx/datashare/services/OutboundEventQueueServiceTest.kt
+++ b/src/test/kotlin/uk/gov/gdx/datashare/services/OutboundEventQueueServiceTest.kt
@@ -53,7 +53,6 @@ class OutboundEventQueueServiceTest {
     unmockkStatic(SqsClient::class)
   }
 
-
   @Test
   fun `sends message`() {
     underTest.sendMessage("queueone", "message")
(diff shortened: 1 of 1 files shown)


src/main/kotlin/uk/gov/gdx/datashare/services/OutboundEventQueueService.kt
```diff
@@ -10,16 +21,240 @@ import uk.gov.gdx.datashare.queue.SqsProperties
...
+
+  private fun isFifoQueue(queueName: String) = queueName.endsWith(".fifo")
+
+  private fun dlqName(queueName: String) =
+    if (isFifoQueue(queueName)) "${queueName.substringBefore(".fifo")}_dlq.fifo" else "${queueName}_dlq"
+
+  private fun keyPolicy(acquirerPrincipal: String?): String {
+    if (acquirerPrincipal == null) {
+      return """
+        {
```
oskarwilliams, line 139: Is there a nicer way of doing this by stringifying a data class or something?

ethanmills: It depends what you think is nicer. It think that would look something like 
```kotlin
@JsonInclude(JsonInclude.Include.NON_NULL)
data class QueuePolicy(
  private val Version: String = "2012-10-12",
  private val Id: String = "cross-account-access",
  private val Statement: List<QueuePolicyStatment>,
)

data class QueuePolicyStatment(
  private val Sid: String,
  private val Effect: String,
  private val Action: List<String>,
  private val Principal: PolicyPrincipal? = null,
  private val Condition: PolicyCondition? = null,
)

data class PolicyPrincipal(
  private val AWS: String,
)

data class PolicyCondition(
  private val Bool: BoolPolicyCondition,
)

data class BoolPolicyCondition(
  @JsonProperty("aws:SecureTransport")
  private val secureTransport: String
)

fun newQueuePolicy(acquirerPrincipal: String?): String {
  if (acquirerPrincipal == null) {
    return QueuePolicy(
      Statement = listOf(
        QueuePolicyStatment(
          Sid = "httpsonly",
          Effect = "Deny",
          Action = listOf("sqs:*"),
          Condition = PolicyCondition(
            BoolPolicyCondition(secureTransport = "false")
          )
        )
      )
    ).toJson()
  }
}
```
which is much less readable to my mind

oskarwilliams: I thought there may be an existing type in the sdk, but if not I am happy to go with as is

ethanmills: No types - everything just takes a string policy document.
```kotlin
mapOf(
  "Version" to "2012-10-12",
  "Id" to "cross-account-access",
  "Statement" to listOf(
    mapOf(
      "Sid" to "acquirerAccess",
      "Principal" to mapOf("AWS" to "$acquirerPrincpal"),
      "Effect" to "Allow",
      "Action" to listOf(
        "SQS:GetQueueUrl",
        "SQS:GetQueueAttributes",
        "SQS:ChangeMessageVisibility",
        "SQS:ReceiveMessage",
        "SQS:DeleteMessage",
      ),
    ),
    mapOf(
      "Sid" to "httponly",
      "Effect" to "Deny",
      "Action" to "sqs:*",
      "Condition" to mapOf("Bool" to mapOf("aws:SecureTransport" to "false"))
    ),
  )
)
```
is a little better than the above, but I think I still prefer the hand-rolled JSON

oskarwilliams (approved):

PR #1213, opened 2023-05-24
GPC-383: API ability to create acquirer queues

ethanmills:
I've not made the change to /admin in this one to keep the review smaller

### DIFF
--- a/src/main/kotlin/uk/gov/gdx/datashare/models/validators/SingleConsumptionMethodValidator.kt
+++ b/src/main/kotlin/uk/gov/gdx/datashare/models/validators/SingleConsumptionMethodValidator.kt
@@ -1,0 +1,29 @@
+package uk.gov.gdx.datashare.models.validators
+
+import jakarta.validation.ConstraintValidator
+import jakarta.validation.ConstraintValidatorContext
+import uk.gov.gdx.datashare.models.AcquirerSubRequest
+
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/gdx/datashare/models/validators/SingleConsumptionMethodValidator.kt
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    if (value.oauthClientId != null && value.queueName != null) {
+      return false
+    }
+
+    if ((value.queueName == null && value.principalArn != null) ||
+      (value.queueName != null && value.principalArn == null)
+    ) {
+      context.disableDefaultConstraintViolation()
```
olevett, line 20: This constraint + the naming in the annotation feels a little bit inconsistent, can I double check what the overall aim of it is?

olevett: I _think_ we're saying you must give us one and only one of
- oauth client ID
- queue name+principal arn
Is that right?

ethanmills: Yes. I'm splitting this out into two constraints and validators - it's really two concepts. If you're happy for it to come in a followup with the /admin stuff I'll do it there, if not I'll change this

PR #1214, opened 2023-05-24
GPC-262: Add failure notification for codedeploy

CarlyG55:


ethanmills:
I suspect we'll also need to make some changes to allow codestar to publish to SNS. Have you seen this? https://docs.aws.amazon.com/dtconsole/latest/userguide/set-up-sns.html

oskarwilliams (approved):

PR #1219, opened 2023-05-25
GPC-297 bastion tunnel notification

CarlyG55:


terraform/modules/sns/main.tf
```diff
@@ -64,11 +64,34 @@ data "aws_iam_policy_document" "kms_codestar_access" {
...
+      type        = "Service"
+      identifiers = ["events.amazonaws.com"]
+    }
+    resources = ["*"]
+    condition {
+      test     = "StringEquals"
+      variable = "aws:ResourceAccount"
+
+      values = [var.account_id]
+    }
```
CarlyG55, line 85: I'm not entirely sure on this bit as I couldn't find docs, so copied from the S3 one

CarlyG55:
Currently have notifs turned on for dev and demo, not set up for prod

terraform/modules/data_share_service/eventbridge.tf
```diff
@@ -0,0 +1,19 @@
+resource "aws_cloudwatch_event_rule" "tunnel_notification" {
+  name        = "bastion-tunnel-notification"
+  description = "Send an SNS notification when a tunnel is initiated to the bastion host"
+
+  event_pattern = jsonencode({
```
oskarwilliams, line 5: So this is just starting a session, do we want to have anything about where the tunnel is going?

CarlyG55: Yeah I originally wasn't sure how this is done, but I've had a further dig into the docs & cloudtrail logs and think I might have found a way to do it

oskarwilliams (approved):

PR #1240, opened 2023-05-30
GPC-235: add role assumptions to prod

olevett:


olevett:
🤦  it does indeed already exist, I misread one of the errors trying to test the alert and jumped to completely the wrong conclusion.

PR #1253, opened 2023-05-31
GPC-364 Add event generation to fake gro

CarlyG55:


terraform/modules/gro_ingestion_service/lambdas/src/functions/insertXml.ts
```diff
@@ -0,0 +1,127 @@
...
+                PersonNameSuffix: "",
+            },
+            DeceasedAliasNameType: "",
+            DeceasedMaidenNameType: "",
+            DeceasedGender: getRandomElement(genderOptions),
+            DeceasedDeathDate: {
+                PersonDeathDate: getRandomElement(dateOfDeathOptions),
+                VerificationLevel: "03",
+            },
+            PartialMonthOfDeath: "",
```
olevett, line 100: Should we be populating these fields from the DeathDate?

terraform/modules/gro_ingestion_service/lambdas/src/functions/insertXml.ts
```diff
@@ -0,0 +1,127 @@
...
+    ["10 Test Street", "Test Town", "Test County"],
+    ["", "Westminster", "London"],
+    ["27 Brown Lane", "", "Somerset"],
+    ["65 Link Road", "Southport", ""],
+]
+const postcodeOptions = ["PR8 1HY", ""]
+
+const handler: Handler = async (event): Promise<InsertXmlResponse> => {
+    const numberOfRecords =
+        event["detail-type"] === "Scheduled Event" || !event["numberOfRecords"] ? 25 : event["numberOfRecords"]
```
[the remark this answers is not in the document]
CarlyG55: I'm pretty sure it's a default - I checked through the len scheduled lambda and couldn't find anything, and this is what that uses

terraform/modules/gro_ingestion_service/lambdas/src/functions/insertXml.ts
CarlyG55: I filled in the sections that were filled in in the example schema - [here](https://docs.google.com/spreadsheets/d/1fFMlFL52P2FYAd5X-0PlGbKuPn-4uEkn/edit#gid=1526528896) someone seemed to ask if these should be calculated, but that seems to be on the DWP end?

oskarwilliams: So at the moment we are passing this through so I would suggest adding it, once we get the sample JSON back from DWP we will have to update all our models so thats ok to update then

terraform/demo/main.tf
```diff
@@ -116,4 +116,6 @@ module "gro_ingestion_service" {
   auth_url                = module.data_share_service.token_auth_url
   publisher_client_id     = module.data_share_service.gro_ingestion_client_id
   publisher_client_secret = module.data_share_service.gro_ingestion_client_secret
+
+  insert_xml_lambda_schedule = "rate(5 minutes)"
```
oskarwilliams, line 120: could we do 1 hours for the moment in both envs, we can update to 5 minutes when these get actively used

terraform/modules/gro_ingestion_service/lambdas/src/functions/insertXml.ts
```diff
@@ -0,0 +1,129 @@
...
+    const putObjectCommand = new PutObjectCommand({
+        Bucket: config.s3BucketArn,
+        Key: fileKey,
+        Body: xml,
+    })
+    await s3Client.send(putObjectCommand)
+}
+
+const generateXml = (numberOfRecords: number) => {
+    const deathRecords = [...Array(numberOfRecords)].map(_ => createDeathRecord())
```
oskarwilliams, line 60:
```suggestion
    const deathRecords = [...Array(numberOfRecords)].map(createDeathRecord)
```

I _think_ you don't need the `_ => `

oskarwilliams (changes requested):

oskarwilliams (approved):

PR #1255, opened 2023-05-31
GPC-383: fix sqs policy

ethanmills:


src/main/kotlin/uk/gov/gdx/datashare/services/OutboundEventQueueService.kt
ethanmills, line 1: I have no idea how I did this

oskarwilliams (approved):

PR #1268, opened 2023-06-01
GPC-NONE: Fix some admin delete schema

oskarwilliams:


src/main/kotlin/uk/gov/gdx/datashare/controllers/AcquirersController.kt
```diff
@@ -92,7 +92,7 @@ class AcquirersController(
     description = "Mark an acquirer as deleted. This will also soft delete all acquirer subscriptions linked to this acquirer, and delete the queues or cognito clients for those subscriptions, if no more subscriptions are attached to the queue or cognito client.",
     responses = [
       ApiResponse(
-        responseCode = "201",
+        responseCode = "204",
```
olevett, line 95: I assume we are actually getting a 204 from this API?

oskarwilliams: Yep

olevett (approved):

PR #1271, opened 2023-06-02
GPC-NONE: Upgrade to Gradle 8.1.1

olevett:


olevett:
Draft because https://github.com/springdoc/springdoc-openapi-gradle-plugin/issues/107 - looking at a couple of workarounds

PR #1272, opened 2023-06-02
GPC-NONE: Fix spring security deprecations

ethanmills:
Please review carefully. Per https://docs.spring.io/spring-security/reference/servlet/configuration/java.html#_multiple_httpsecurity_instances (kotlin docs just seem to be wrong)

### DIFF
--- a/src/main/kotlin/uk/gov/gdx/datashare/config/ResourceServerConfiguration.kt
+++ b/src/main/kotlin/uk/gov/gdx/datashare/config/ResourceServerConfiguration.kt
@@ -4,6 +4,7 @@ import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
+import org.springframework.core.Ordered
 import org.springframework.core.annotation.Order
 import org.springframework.security.config.annotation.authentication.builders.AuthenticationManagerBuilder
 import org.springframework.security.config.annotation.method.configuration.EnableMethodSecurity
@@ -65,6 +66,7 @@ class ResourceServerConfiguration {
       .build()
   }
 
+  @Order(Ordered.LOWEST_PRECEDENCE)
   @Bean
   fun apiFilterChain(http: HttpSecurity): SecurityFilterChain {
     return http
--- a/src/test/kotlin/uk/gov/gdx/datashare/integration/actuator/HealthCheckTest.kt
+++ b/src/test/kotlin/uk/gov/gdx/datashare/integration/actuator/HealthCheckTest.kt
@@ -1,4 +1,4 @@
-package uk.gov.gdx.datashare.integration.health
+package uk.gov.gdx.datashare.integration.actuator
 
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Test
--- a/src/test/kotlin/uk/gov/gdx/datashare/integration/actuator/InfoTest.kt
+++ b/src/test/kotlin/uk/gov/gdx/datashare/integration/actuator/InfoTest.kt
@@ -1,4 +1,4 @@
-package uk.gov.gdx.datashare.integration.health
+package uk.gov.gdx.datashare.integration.actuator
 
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Test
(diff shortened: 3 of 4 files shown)


src/main/kotlin/uk/gov/gdx/datashare/config/ResourceServerConfiguration.kt
```diff
@@ -30,26 +31,48 @@ class ResourceServerConfiguration {
       .roles(role)
   }
 
+  @Order(1)
```
olevett, line 34: I got part way through this, could we make sure we've got a test for it too?

olevett: As a really quick starter, I think something like this:
```kotlin
package uk.gov.gdx.datashare.uk.gov.gdx.datashare.integration.health

import org.junit.jupiter.api.Test
import uk.gov.gdx.datashare.integration.SqsIntegrationTestBase

class PrometheusTest : SqsIntegrationTestBase() {
  @Test
  fun `Unauthenticated page is inaccessible`() {
    webTestClient.get()
      .uri("/prometheus")
      .exchange()
      .expectStatus()
      .isUnauthorized
  }

  @Test
  fun `Authenticated page is accessible`() {
    webTestClient
      .get()
      .uri("/prometheus")
      .headers { it.setBasicAuth("prometheus", "prometheus") }
      .exchange()
      .expectStatus()
      .isOk
  }
}
```

oskarwilliams (commented):
If we can add some tests that would be great, but it all looks accurate to me.

src/main/kotlin/uk/gov/gdx/datashare/config/ResourceServerConfiguration.kt
```diff
@@ -30,26 +31,48 @@ class ResourceServerConfiguration {
...
+      .securityMatcher("/prometheus")
+      .authorizeHttpRequests {
+        it.anyRequest().hasRole("ACTUATOR")
+      }
+      .httpBasic {}
+      .build()
+  }
+
+  @Bean
+  fun apiFilterChain(http: HttpSecurity): SecurityFilterChain {
```
olevett, line 69: Does no Order on this matter?

src/main/kotlin/uk/gov/gdx/datashare/config/ResourceServerConfiguration.kt
```diff
@@ -30,26 +31,48 @@ class ResourceServerConfiguration {
...
+  }
+
+  @Order(2)
+  @Bean
+  fun prometheusFilterChain(http: HttpSecurity): SecurityFilterChain {
+    return http
+      .csrf { it.disable() }
+      .securityMatcher("/prometheus")
+      .authorizeHttpRequests {
+        it.anyRequest().hasRole("ACTUATOR")
```
olevett, line 62: Whilst we're here, we hard code the role on this line, but the role is configurable for the user itself - I wonder if we should either configure this one, or hard code the other role?

src/main/kotlin/uk/gov/gdx/datashare/config/ResourceServerConfiguration.kt
ethanmills: Was just copied from the docs - happy to add LAST

src/main/kotlin/uk/gov/gdx/datashare/config/ResourceServerConfiguration.kt
ethanmills: Yeah wasn't sure what that was about - @oskarwilliams  what was the intention?

oskarwilliams (approved):

src/main/kotlin/uk/gov/gdx/datashare/config/ResourceServerConfiguration.kt
ethanmills: It's `Ordered.LOWEST_PRECEDENCE` which is `Integer.MAX_VALUE` under the hood 😄

src/main/kotlin/uk/gov/gdx/datashare/config/ResourceServerConfiguration.kt
ethanmills: I'm looking to fiddle with the actuator stuff a bit, but will do that in a separate change. Will leave this until then

olevett (approved):

PR #1283, opened 2023-06-05
GPC-NONE: Add key deletion to queue deletion

oskarwilliams:


terraform/modules/data_share_service/iam.tf
```diff
@@ -213,9 +213,10 @@ data "aws_iam_policy_document" "ecs_task_manage_acquirer_queues" {
   }
 
   statement {
-    sid = "createNewKeys"
+    sid = "createAndDeleteKeys"
     actions = [
-      "kms:CreateKey"
+      "kms:CreateKey",
+      "kms:ScheduleKeyDeletion"
```
ethanmills, line 219: I think we only want to allow deletion of the correct keys. Something like this: https://docs.aws.amazon.com/kms/latest/developerguide/conditions-kms.html#conditions-kms-resource-aliases:~:text=The%20following%20example%20IAM%20policy%20statement%20allows%20the%20principal%20to%20enable%20and%20disable%20KMS%20keys%20but%20only%20when%20all%20aliases%20of%20the%20KMS%20keys%20include%20%22Test.%22

ethanmills (changes requested):

terraform/modules/data_share_service/iam.tf
oskarwilliams: Should have fixed to this

src/main/kotlin/uk/gov/gdx/datashare/services/OutboundEventQueueService.kt
```diff
@@ -269,11 +281,42 @@ class OutboundEventQueueService(
...
 
+  private fun deleteQueue(queueName: String) {
+    val queue = getQueue(queueName)
+    val deleteQueueRequest = DeleteQueueRequest.builder().queueUrl(queue.queueUrl).build()
+    queue.sqsClient.deleteQueue(deleteQueueRequest)
+  }
+
+  private fun deleteKeyForQueue(queueName: String) {
+    val alias = generateAlias(queueName)
+    val keyId = getKeyAlias(alias).targetKeyId()
```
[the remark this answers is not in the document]
oskarwilliams: After discussion, unsurprisingly you are correct, getting from the attributes is best :)

src/test/kotlin/uk/gov/gdx/datashare/services/AcquirersServiceTest.kt
```diff
@@ -316,8 +316,8 @@ class AcquirersServiceTest {
 
     every { adminActionAlertsService.noticeAction(any()) } just runs
 
-    every { outboundEventQueueService.deleteQueue(queueAcquirerSubscription.queueName!!) } just runs
-    every { outboundEventQueueService.deleteQueue("${queueAcquirerSubscription.queueName}_dlq") } just runs
+    every { outboundEventQueueService.deleteAcquirerQueueAndDlq(queueAcquirerSubscription.queueName!!) } just runs
+    every { outboundEventQueueService.deleteAcquirerQueueAndDlq("${queueAcquirerSubscription.queueName}_dlq") } just runs
```
ethanmills, line 320: Do we need this?

ethanmills (approved):
I think there are a few unnecessary mocks in there?

PR #1287, opened 2023-06-06
GPC-198: Add CloudWatch alarms to SQS queues

oskarwilliams:


olevett:
From some discussion yesterday, I think we're changing tack a little, and getting Spring to kick out queue metrics and then building out reporting/alarms in Grafana/Prometheus

PR #1295, opened 2023-06-07
GPC-300: Restrict VPC ACLs

ethanmills:
There's a chance that this will break something, but it should be obvious fairly quickly

### DIFF
--- a/terraform/modules/data_share_service/vpc.tf
+++ b/terraform/modules/data_share_service/vpc.tf
@@ -21,12 +21,22 @@ locals {
       action     = "ALLOW"
       protocol   = "tcp"
     },
-    # Allow response to outbound calls
+    # Deny port 3389 (RDP)
     {
       rule_no    = 3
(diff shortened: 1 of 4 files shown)


terraform/modules/data_share_service/vpc.tf
```diff
@@ -1,3 +1,167 @@
...
+      cidr_block = "0.0.0.0/0"
+      action     = "DENY"
+      protocol   = -1
+    }
+  ]
+  private_acl_egress_rules = [
+    # Allow traffic to database from ECS service
+    {
+      rule_no    = 1
+      from_port  = 45678
```
olevett, line 55: This presumably has to match a few places (e.g. `rds_bastion_host_sg`, `rds_postgres_cluster`) is it worth pulling it out as a variable rather than just hard coding the same number in a few places?

olevett (approved):

PR #1310, opened 2023-06-08
GPC-198: Fix queue metrics

oskarwilliams:


src/main/kotlin/uk/gov/gdx/datashare/services/ScheduledJobService.kt
```diff
@@ -38,14 +39,21 @@ class ScheduledJobService(
...
+    queueMetrics.forEach { (queueName, queueMetric) ->
+      val (ageMeter, dlqLengthMeter) = queueMetersMap.computeIfAbsent(queueName) {
+        Pair(
+          meterRegistry.gauge("age_of_oldest_message", listOf(Tag.of("queue_name", it)), AtomicInteger(0))!!,
+          meterRegistry.gauge("dlq_length", listOf(Tag.of("queue_name", it)), AtomicInteger(0))!!,
+        )
+      }
+      ageMeter.set(queueMetric.ageOfOldestMessage ?: 0)
+      if (queueMetric.ageOfOldestMessage == null) {
+        log.error("No age_of_oldest_message found for queue: $queueName")
```
olevett, line 51: What should we do in this case? If we're not going to act, I think I'd be tempted to make it a warning rather than an error

oskarwilliams: Yep fair argument, will do

PR #1327, opened 2023-06-09
GPC-NONE: add redundancy to ssm endpoint

ethanmills:


terraform/modules/data_share_service/ec2.tf
ethanmills, line 1: I don't think the other services are critical in the same way here - they're only used by the bastion host which is in subnet 0, so there's no point in redundancy

oskarwilliams (approved):

PR #1332, opened 2023-06-09
GPC-NONE: Tag acquirer name

oskarwilliams:


ethanmills (approved):
This is fine at this scale. It's an N+1 query, but it's one we're fine with at the moment

PR #1333, opened 2023-06-09
GPC-NONE: Update remove meters

oskarwilliams:


ethanmills (approved):
Why this change? Is it so we don't mutate the list as we map through it?

oskarwilliams:
It errors if we mutate the list as we map through it

PR #1383, opened 2023-06-15
GPC-272: IP restrict admin access

ethanmills:


terraform/dev/main.tf
```diff
@@ -106,6 +106,8 @@ module "data_share_service" {
 
   admin_alerts_enabled           = false
   database_tunnel_alerts_enabled = false
+
+  admin_login_allowed_ip_blocks = null
```
oskarwilliams, line 110: can we not just provide a default and not set it here?

Mjwillis (approved):
LGTM

oskarwilliams (approved):

terraform/dev/main.tf
ethanmills: I'd like not restricting to be a conscious choice so we don't forget it in prod

terraform/demo/main.tf
```diff
@@ -106,6 +106,25 @@ module "data_share_service" {
 
   admin_alerts_enabled           = false
   database_tunnel_alerts_enabled = false
+
+  admin_login_allowed_ip_blocks = [
```
olevett, line 110: I don't know if this definitely true of the resources, but various of them end up doing a sort at some point, so we might end up with unexpected changes if we don't order this - I'd be tempted to apply it and then see if a plan ends recreating any

ethanmills: I think this provider is slightly more intelligent: https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/wafv2/ip_set.go#L68

ethanmills: Given ☝️ I'm confident enough to merge this and test afterwards, fixing if necessary

PR #1386, opened 2023-06-15
GPC-371: Use service domain

ethanmills:
This change has the following effects:

- Change cognito user pool admin client permitted callback url to the `service.gov.uk` version
- Change `SPRINGDOC_SWAGGER_UI_OAUTH2_REDIRECT_URL` env var to `service.gov.uk` version
- Change len lambdas to point at `service.gov.uk` domain
- Change gro lambdas to point as `service.gov.uk` domain


oskarwilliams (approved):

olevett:
Does this have any negative effect on DWP? Can you still auth through the auto generated cognito urls?

PR #1387, opened 2023-06-15
GPC-190 Investigate and document our processes for dealing with a deployment we want to rollback

Mjwillis:


doc/runbooks/rollback-application-releases.md
oskarwilliams, line 1: This feels a lot more like an ADR than a runbook of what to do. Could you maybe slightly reformat this as an ADR?

For a runbook, we probably want to say what to do and how to do it. So in this case, the rollback process is
1. Code release rollback
 - Go into the CodeDeploy in the AWS console. 
 - Find the previous deployment and click into it
 - Click retry deployment, this will cause the deployed version to be rolled back
 - Fix and merge new code into git
2. DB migration rollback
 - Commit new code that fixes the db
 - Merge and release

The DB migration rollback is actually a rollforward as once some migrations have been applied to our db it would be better to write a new migration to fix it rather than anything else

oskarwilliams (changes requested):

doc/runbooks/rollback-application-releases.md
Mjwillis: Thanks @oskarwilliams I've made changes as suggested

oskarwilliams (commented):
One small change, I'm gonna let @ethanmills / @olevett have a look as well :)

doc/architecture/decisions/0027-recovery-rollback-strategy.md
```diff
@@ -0,0 +1,97 @@
...
+
+#### Advantages
+Fast and easy to execute.
+
+#### Disadvantages
+Can cause data loss, inconsistency, or corruption if the new version has made any changes to the database or other external resources.
+
+
+# Decision
+Although larger and more complex systems may benefit from blue/green deployment due to its reduced risk of errors and data issues, more frequent and urgent deployments favour deployment rollback, as it is quicker and simpler to execute. The simplicity of rollback was chosen as the approach.
```
ethanmills, line 71: It's worth noting that we _do_ have blue/green deployment, with automated smoke tests. We could easily configure this for something other than all at once deployment: https://github.com/alphagov/gdx-data-share-poc/blob/main/terraform/modules/data_share_service/codedeploy.tf#L10.

We could also configure the blue instances to stick around for faster rollback: https://github.com/alphagov/gdx-data-share-poc/blob/main/terraform/modules/data_share_service/codedeploy.tf#L23

I think the approach here is correct for once the blue/green phase is over, or for db migration issues

doc/runbooks/application-recovery.md
```diff
@@ -0,0 +1,21 @@
...
+
+### Code release rollback
+
+Use these steps to roll back to the previous application version:-
+
+1. Go into the CodeDeploy in the AWS console.
+2. Find the previous deployment and click into it.
+3. Click retry deployment, this will cause the deployed version to be rolled back.
+4. Fix and merge new code into git.
+
```
ethanmills, line 13: Have we done a trial run of this?

ethanmills (approved):

doc/runbooks/application-recovery.md
oskarwilliams: I have yep, or at least the redeployment of old code

PR #1398, opened 2023-06-16
GPC-NONE: Rename GitHub Repo

olevett:


src/main/kotlin/uk/gov/gdx/datashare/services/OutboundEventQueueService.kt
```diff
@@ -116,7 +116,7 @@ class OutboundEventQueueService(
         Tag.builder().tagKey("Environment").tagValue(environment).build(),
         Tag.builder().tagKey("Owner").tagValue("gdx-dev-team@digital.cabinet-office.gov.uk").build(),
         Tag.builder().tagKey("Product").tagValue("Government Data Exchange").build(),
-        Tag.builder().tagKey("Repository").tagValue("https://github.com/alphagov/gdx-data-share-poc").build(),
+        Tag.builder().tagKey("Repository").tagValue("https://github.com/alphagov/di-data-life-events-platform").build(),
```
ethanmills, line 119: I think this will fail a test

olevett: We'll also possibly want to go and amend any queues we've got already created

ethanmills (approved):
I have no idea what is wrong with shared TF. I also think this might break techdocs, but I'll fix that. The URL will change though

PR #1426, opened 2023-06-21
GPC-NONE: Update dependabot for terraform and platform dockerfile

oskarwilliams:


olevett (approved):
It's depressing that dependabot doesn't have a better way to handle this, but I haven't found one from a dive!

PR #1448, opened 2023-06-21
GPC-NONE: Update Lint to work in new directory structure

olevett:


oskarwilliams (approved):
This is good for now, but it may be **/... later instead

PR #1458, opened 2023-06-22
GPC-NONE: Add test lambda

oskarwilliams:


lambdas/test-lambda/src/test/java/uk/gov/di/data/lep/MaintTest.java
```diff
@@ -0,0 +1,22 @@
+package lep;
```
olevett, line 1: This package seems wrong

lambdas/test-lambda/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+}
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    implementation "com.amazonaws:aws-lambda-java-core:$rootProject.ext.dependencyVersions.awsLambdaJavaCore"
+
+    testImplementation "org.junit.jupiter:junit-jupiter:5.9.3",
```
olevett, line 13: I think I said this on a different review somewhere, but if we're pinning dependency versions anywhere (e.g. line 11) can we pin them the same way everywhere?

lambdas/test-lambda/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+            "org.mockito:mockito-junit-jupiter:5.3.1"
+}
+
+java {
+    sourceCompatibility = JavaVersion.VERSION_17
+    targetCompatibility = JavaVersion.VERSION_17
+}
+
+test {
+    useJUnitPlatform ()
```
olevett, line 23:
```suggestion
    useJUnitPlatform()
```

lambdas/test-lambda/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+}
+
+jacocoTestReport {
+    dependsOn test
+    reports {
+        xml.required.set(true)
+    }
+}
+
+task wrapper(type: Wrapper){
```
olevett, line 34:
```suggestion
task wrapper(type: Wrapper) {
```

olevett (changes requested):

lambdas/build.gradle
```diff
@@ -0,0 +1,7 @@
+plugins {
+    id "java"
+}
+
+allprojects {
+    tasks.register("prepareKotlinBuildScriptModel"){}
```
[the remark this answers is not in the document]
oskarwilliams: Yes, this is because some of the subprojects have kotlin, so it does this on everything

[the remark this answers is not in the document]
olevett: Right, so this is removing the Kotlin task, because it's defined elsewhere but we don't need it here?

lambdas/test-lambda/build.gradle
oskarwilliams: Done

lambdas/build.gradle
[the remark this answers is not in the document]
oskarwilliams: ACtually nope
This turns out was my IDE being an idiot, removed here

olevett (approved):

PR #1460, opened 2023-06-22
Gpc 162 enrichment field event type relationship

CarlyG55:
Hannah's changes from https://github.com/alphagov/di-data-life-events-platform/pull/1459

### DIFF
--- a/platform/src/main/kotlin/uk/gov/gdx/datashare/config/ApiExceptionHandler.kt
+++ b/platform/src/main/kotlin/uk/gov/gdx/datashare/config/ApiExceptionHandler.kt
@@ -113,8 +113,8 @@ class ApiExceptionHandler {
       )
   }
 
-  @ExceptionHandler(EnrichmentFieldsNotValidForEventType::class)
-  fun handleEnrichmentFieldsNotValidForEventType(e: EnrichmentFieldsNotValidForEventType): ResponseEntity<ErrorResponse?>? {
+  @ExceptionHandler(EnrichmentFieldsNotValidForEventTypeException::class)
+  fun handleEnrichmentFieldsNotValidForEventType(e: EnrichmentFieldsNotValidForEventTypeException): ResponseEntity<ErrorResponse?>? {
     log.debug("Invalid enrichment field(s) for the given event type: {}", e.message)
     return ResponseEntity
       .status(BAD_REQUEST)
@@ -244,7 +244,7 @@ class SupplierSubscriptionNotFoundException(message: String) : Exception(message
 
 class NoDataFoundException(message: String) : Exception(message)
 
-class EnrichmentFieldsNotValidForEventType(message: String) : Exception(message)
+class EnrichmentFieldsNotValidForEventTypeException(message: String) : Exception(message)
 
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @Schema(description = "Error Response")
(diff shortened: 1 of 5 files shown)


platform/src/main/kotlin/uk/gov/gdx/datashare/config/ApiExceptionHandler.kt
```diff
@@ -230,6 +244,8 @@ class SupplierSubscriptionNotFoundException(message: String) : Exception(message
 
 class NoDataFoundException(message: String) : Exception(message)
 
+class EnrichmentFieldsNotValidForEventType(message: String) : Exception(message)
```
oskarwilliams, line 247: I think the rest of these exceptions all end in `..Exception`, could you change the name here too please :)

platform/src/main/kotlin/uk/gov/gdx/datashare/enums/EventType.kt
```diff
@@ -6,3 +6,43 @@ enum class EventType {
   LIFE_EVENT,
   TEST_EVENT,
 }
+
+val EventTypeEnrichmentFieldsRelationship = mapOf(
+  EventType.GRO_DEATH_NOTIFICATION to listOf<EnrichmentField>(
+    EnrichmentField.SOURCE_ID,
+    EnrichmentField.SEX,
+    EnrichmentField.DATE_OF_BIRTH,
+    EnrichmentField.REGISTRATION_ID,
```
oskarwilliams, line 15: I think we should have `DATE_OF_DEATH` here as weel

platform/src/test/kotlin/uk/gov/gdx/datashare/services/AcquirersServiceTest.kt
```diff
@@ -205,6 +206,37 @@ class AcquirersServiceTest {
...
+  @Test
+  fun `addAcquirerSubscription validates enrichmentFields for given eventType`() {
+    every { acquirerRepository.findByIdOrNull(acquirer.id) }.returns(acquirer)
+    every { acquirerSubscriptionRepository.save(any()) }.returns(acquirerSubscription)
+    every { outboundEventQueueService.createAcquirerQueue(any(), any()) } returns ""
+    every { adminActionAlertsService.noticeAction(any()) } just runs
+
+    val acquirerSubRequest = AcquirerSubRequest(
+      EventType.TEST_EVENT,
+      oauthClientId = null,
```
oskarwilliams, line 218: Could you set this to a value. We then wouldn't need the stuff on line 213

platform/src/main/kotlin/uk/gov/gdx/datashare/services/AcquirersService.kt
```diff
@@ -73,8 +76,15 @@ class AcquirersService(
 
   private fun addAcquirerSubscriptionEnrichmentFields(
     acquirerSubscriptionId: UUID,
+    eventType: EventType,
     enrichmentFields: List<EnrichmentField>,
   ): List<AcquirerSubscriptionEnrichmentField> {
+    val eventTypeEnrichmentFields = EventTypeEnrichmentFieldsRelationship[eventType]
+    enrichmentFields.forEach {
+      if (!eventTypeEnrichmentFields!!.contains(it)) {
+        throw EnrichmentFieldsNotValidForEventType("Invalid enrichment field(s) for the given event type")
```
oskarwilliams, line 85: could you add in which enrichment field and which event type into the error message. That would involve add the failing enrichment fields to a list and then throwing if that list has any in it.

oskarwilliams (changes requested):

oskarwilliams (approved):

PR #1482, opened 2023-06-27
GPC-NONE: Update team email

oskarwilliams:


platform/src/main/kotlin/uk/gov/gdx/datashare/services/OutboundEventQueueService.kt
```diff
@@ -114,8 +114,8 @@ class OutboundEventQueueService(
       .keyId(keyId)
       .tags(
         Tag.builder().tagKey("Environment").tagValue(environment).build(),
```
olevett, line 116: Do we need to manually change a queue here?

olevett (approved):

PR #1522, opened 2023-07-03
GPC-437: Secure pipelines and control tower adr

oskarwilliams:


ethanmills (commented):
Happy for this, but will wait for someone else to review

doc/architecture/decisions/0028-secure-pipelines-and-control-tower.md
```diff
@@ -0,0 +1,49 @@
...
+
+Proposed
+
+## Context
+
+The existing platform is trying to align with the rest of DI, which uses AWS Control Tower to manage AWS accounts, and
+Secure Pipelines to deploy their systems. As a result, the platform is attempting to align to this flow, while
+also aligning our architecture (as laid out in [ADR 2#](Put link here)). DI also uses an environment per account flow,
+which is different from our approach of an account for all our non-prod environments, and an account for our prod
+environment.
```
galund, line 17: If we decide we aren't going to migrate the existing thing, then this ADR should probably reflect that @oskarwilliams . I still think Control Tower, then pipelines, then rewriting code into Lambdas/TypeScript is the correct order, but either way let's just make sure the ADRs reflect the decision we're actually making :)

doc/architecture/decisions/0028-secure-pipelines-and-control-tower.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Status
+
+Proposed
+
+## Context
+
+The existing platform is trying to align with the rest of DI, which uses AWS Control Tower to manage AWS accounts, and
+Secure Pipelines to deploy their systems. As a result, the platform is attempting to align to this flow, while
+also aligning our architecture (as laid out in [ADR 2#](Put link here)). DI also uses an environment per account flow,
```
olevett, line 15:
```suggestion
also aligning our architecture (as laid out in [ADR 2#](0002-gdx-share-architecture.md)). DI also uses an environment per account flow,
```

olevett (approved):

doc/architecture/decisions/0028-secure-pipelines-and-control-tower.md
oskarwilliams: this will be the ADR for the DI alignment, not 2

doc/architecture/decisions/0028-secure-pipelines-and-control-tower.md
oskarwilliams: So we are going to migrate the existing terraform into secure pipelines, independent of the order we do things (as a lot of this will be parallel), so I will leave this as is

PR #1528, opened 2023-07-04
GPC-440: Data in demo

ethanmills:
DWP will now be using this environment in addition to dev

### DIFF
--- a/terraform/demo/main.tf
+++ b/terraform/demo/main.tf
@@ -127,6 +127,22 @@ module "data_share_service" {
   ]
 }
 
+module "len" {
+  source                      = "../modules/len"
+  environment                 = local.env
+  region                      = data.aws_region.current.name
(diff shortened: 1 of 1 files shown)


terraform/demo/main.tf
```diff
@@ -127,6 +127,22 @@ module "data_share_service" {
   ]
 }
 
+module "len" {
+  source                      = "../modules/len"
+  environment                 = local.env
+  region                      = data.aws_region.current.name
+  schedule                    = "rate(1 minute)"
```
oskarwilliams, line 134: do we want this frequency for test as well, or do they want it more realistic?

oskarwilliams (approved):
One minor comment but approved no matter the outcome of it

terraform/demo/main.tf
ethanmills: Going to start with this and tweak

PR #1546, opened 2023-07-06
Enrichment lambda

hannah-leong:


lib/src/main/java/uk/gov/di/data/lep/library/dto/GroDeathEventBaseData.java
```diff
@@ -0,0 +1,5 @@
+package uk.gov.di.data.lep.library.dto;
+
+public record GroDeathEventBaseData(
+    String sourceId) {
+}
```
oskarwilliams, line 5: I'd rather this be a single line as it is so short :)

oskarwilliams (approved):

PR #1595, opened 2023-07-12
GPC-NONE: Update test to include lambdas

oskarwilliams:


.github/workflows/application-pull-request.yml
```diff
@@ -24,6 +24,8 @@ concurrency:
 jobs:
   test:
     uses: ./.github/workflows/test.yml
+    secrets:
+      SAM_APP_VALIDATE_ROLE_ARN: ${{ secrets.SAM_APP_VALIDATE_ROLE_ARN }}
```
olevett, line 28: Is this really that sensitive? Or does it differ by env?

.github/workflows/test.yml
```diff
@@ -24,28 +27,25 @@ jobs:
...
-      - name: Report
-        uses: dorny/test-reporter@v1
-        if: always()
-        with:
-          name: Test results
-          path: ${{ github.workspace }}/platform/build/test-results/test/TEST-*.xml
-          reporter: java-junit
-          only-summary: true
-
-      - name: Report test coverage
```
olevett, line 38: Do we not want test coverage?

.github/workflows/test.yml
```diff
@@ -24,28 +27,25 @@ jobs:
         uses: gradle/gradle-build-action@v2.6.0
 
       - name: Execute Gradle build
-        run: ./gradlew clean platform:test
+        run: ./gradlew clean test
 
-      - name: Report
```
olevett, line 29: @ethanmills  I think you added this test reporter, is it better/different to the mikepenz one?

olevett (approved):

.github/workflows/application-pull-request.yml
oskarwilliams: This is from the guide from DI, so I have kept it sensitive for their benefit

.github/workflows/test.yml
oskarwilliams: We have that with a separate reporter, conforming with DI

oskarwilliams: Reverting this

.github/workflows/test.yml
oskarwilliams: he's happy for us to try this

PR #1596, opened 2023-07-12
GPC-NONE: Add initial secure pipeline dependencies stack bash

oskarwilliams:


scripts/initialise_secure_pipelines_dependencies_stack.sh
```diff
@@ -0,0 +1,20 @@
+aws cloudformation create-stack --stack-name github-identity \
```
olevett, line 1: How do I use this and what is it for?

olevett (approved):

PR #1604, opened 2023-07-13
GPC-465: Add logging to the lambdas

hannah-leong:


lambdas/death-enrichment/src/main/java/uk/gov/di/data/lep/GroDeathEnrichment.java
```diff
@@ -16,11 +16,13 @@ public class GroDeathEnrichment
     implements RequestHandler<GroDeathEventBaseData, GroDeathEventEnrichedData> {
     @Override
     public GroDeathEventEnrichedData handleRequest(GroDeathEventBaseData baseData, Context context) {
+        logger = context.getLogger();
         var enrichedData = enrichData(baseData);
         return publish(enrichedData);
     }
 
     public GroDeathEventEnrichedData enrichData(GroDeathEventBaseData baseData) {
+        logger.log("Enriching data (sourceId: " + baseData.sourceId() + ")");
```
oskarwilliams, line 25: do we need to specify log levels?

lambdas/death-enrichment/src/test/java/uk/gov/di/data/lep/GroDeathEnrichmentTest.java
```diff
@@ -25,6 +33,8 @@ void enrichGroDeathEventDataReturnsEnrichedData() {
 
         var result = underTest.handleRequest(baseData, context);
 
+        verify(logger).log("Enriching data (sourceId: " + baseData.sourceId() + ")");
```
oskarwilliams, line 36: I'd rather we pass in the actual string rather than create it

lambdas/death-minimisation/src/test/java/uk/gov/di/data/lep/GroDeathNotificationMinimisationTest.java
```diff
@@ -69,6 +78,7 @@ void minimiseGroDeathEventDataReturnsMinimisedDataWithNoEnrichmentFields() {
 
     @Test
     void minimiseGroDeathEventDataReturnsMinimisedDataWithPartialEnrichmentFields() {
+        when(context.getLogger()).thenReturn(logger);
```
oskarwilliams, line 81: I think we should put this in a before all, and then we can verify its called on all of these easily

lambdas/death-validation/src/main/java/uk/gov/di/data/lep/GroDeathValidation.java
```diff
@@ -15,12 +15,15 @@ public class GroDeathValidation
...
         publish(event);
         return new APIGatewayProxyResponseEvent().withStatusCode(201);
     }
 
     private GroDeathEventBaseData validateRequest(APIGatewayProxyRequestEvent event) {
+        logger.log("Validating request");
+
         GroDeathEvent groDeathEvent;
         try {
             groDeathEvent = new ObjectMapper().readValue(event.getBody(), GroDeathEvent.class);
```
oskarwilliams, line 29: could we add a log below here in the catch and in the failure case saying something like failed to validate event

lib/src/main/java/uk/gov/di/data/lep/library/LambdaHandler.java
```diff
@@ -16,9 +18,11 @@ public O publish(O output) {
         }
```
oskarwilliams, line 18: could we log an error if it fails to process

oskarwilliams (changes requested):

oskarwilliams (approved):

PR #1616, opened 2023-07-14
GPC-NONE: Add 0 reserved concurrent executions

oskarwilliams:


ethanmills:
What's this for?

ethanmills:
This can be set globally and overridden if needed https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-specification-template-anatomy-globals.html

ethanmills:
Doesn't reserved concurrency of zero mean the function can never run? Or does 0 have a special meaning?

oskarwilliams:
Ahh yes, I was treating reserved concurrency as the prewarmed amount which is wrong, I need to think of a better value for this

PR #1630, opened 2023-07-17
GPC-NONE: Add powertools tracing

oskarwilliams:


lib/build.gradle
```diff
@@ -1,18 +1,23 @@
...
     id "jacoco"
+    id "io.freefair.aspectj.post-compile-weaving" version "6.6.3"
 }
 
 repositories {
     mavenCentral()
 }
 
 dependencies {
+
```
ethanmills, line 12: whitespace?

lambdas/authorisation/src/main/java/uk/gov/di/data/lep/Authorisation.java
```diff
@@ -2,10 +2,14 @@
 
 import com.amazonaws.services.lambda.runtime.Context;
 import com.amazonaws.services.lambda.runtime.RequestHandler;
+import software.amazon.lambda.powertools.logging.Logging;
+import software.amazon.lambda.powertools.tracing.Tracing;
 
-public class Authorisation implements RequestHandler {
+public class Authorisation implements RequestHandler<Object, Object> {
```
ethanmills, line 8: Is this change needed? Can we be any more specific?

lib/build.gradle
oskarwilliams: sad

lambdas/authorisation/src/main/java/uk/gov/di/data/lep/Authorisation.java
oskarwilliams: This is to stop it shouting at me, but we don't need it (it is a minor tidy)

lib/build.gradle
oskarwilliams:
```suggestion
```

PR #1686, opened 2023-07-20
GPC-NONE: Remove flaky test

oskarwilliams:


ethanmills (commented):
This seems like an important thing to test

PR #1689, opened 2023-07-20
GPC-NONE: Update spring framework boot

oskarwilliams:


olevett (approved):

olevett:
Could we remove the trivy ignore at the same time?

PR #1708, opened 2023-07-24
GPC-NONE: Update for checkov tagging and vpc stack name

oskarwilliams:


template.yaml
```diff
@@ -187,6 +187,8 @@ Resources:
           "responseLatency":"$context.responseLatency",
           "responseLength":"$context.responseLength"
           }
+      Tags:
+        CheckovRulesToSkip: "CKV_AWS_120"
```
ethanmills, line 191: Are you able to add comments with the what and why?

oskarwilliams: There is a comment already, line 158 for this one. The comment cannot be in the tag however :(

ethanmills (approved):

PR #1710, opened 2023-07-24
GPC-NONE: Domain outputs

ethanmills:


manualStacks/dns/README.md
```diff
@@ -33,33 +33,33 @@ Replace `<environment>` with `dev`, `build`, `staging`, `integration`, `producti
 
 ### Creating a New Stack
 
-Set your AWS profile to the correct environment, the run:
+Set your AWS profile to the correct environment, run `export ENVIRONMENT=<environment>` then run:
```
oskarwilliams, line 36: You also need to log in to the correct AWS profile for running the command, is this confusing that as well?

ethanmills: I intended it read as a list of instructions. I'd be happy to add an oxford comma or take other suggestions

oskarwilliams: yeah could you add this pls 
```suggestion
Set your AWS profile to the correct environment, then run `export ENVIRONMENT=<environment>`, then run:
```

oskarwilliams (approved):

PR #1793, opened 2023-07-31
GPC-512: Add authorisation logic

hannah-leong:


lambdas/authorisation/src/test/java/uk/gov/di/data/lep/AuthorisationTest.java
oskarwilliams, line 1: There are probably a few tests I'd like to add:
* testing we are properly generating the ARN
* testing we are properly setting the subject
* testing what happens if there isn't a scope claim set on the token

oskarwilliams (changes requested):

lambdas/authorisation/src/main/java/uk/gov/di/data/lep/Authorisation.java
```diff
@@ -31,32 +37,34 @@ public Authorisation(Config config, JwtService jwtService) {
...
-        }
-
-        var eventContext = new HashMap<String, String>();
-        eventContext.put("sub", sub);
 
+        var decodedJwt = getDecodedJwt(event.getHeaders());
         var proxyContext = event.getRequestContext();
         var identity = proxyContext.getIdentity();
-
+        var auth = isAllowedPath(decodedJwt, event) ? "Allow" : "Deny";
```
oskarwilliams, line 45:
```suggestion

        var auth = isAllowedPath(decodedJwt, event) ? "Allow" : "Deny";
```

oskarwilliams (approved):
One minor comment

PR #1794, opened 2023-07-31
GPC-492: Add mapping to step function

oskarwilliams:
This adds the map section, and its required S3 upload, to the step function.

The desired flow of the step function will be:
1. S3 action to get object that triggered step function, passes into
2. Lambda to convert object into JSON, passes into
3. S3 action to put object into S3
4. Map action that uses the object in S3 from above
5. Many publish lambdas to send data to LEP
6. If Map succeeds, delete original object
7. If Map succeeds or fails, delete object created in 3
8. If Map fails, alert

Before PR | After PR
----------|-------
image | image


template.yaml
```diff
@@ -740,20 +778,69 @@ Resources:
           - Fn::ImportValue: !Sub ${VpcStackName}-PrivateSubnetIdB
           - Fn::ImportValue: !Sub ${VpcStackName}-PrivateSubnetIdC
       Tags:
-        CheckovRulesToSkip: "CKV_AWS_115.CKV_AWS_117.CKV_AWS_173"
+        CheckovRulesToSkip: "CKV_AWS_115.CKV_AWS_116.CKV_AWS_117.CKV_AWS_173"
+
+  GroConvertToJsonFunctionLogGroup:
+    Type: AWS::Logs::LogGroup
+    Properties:
+      RetentionInDays: 14
```
ethanmills, line 786: We probably need to have a global think about log retention at some point. 14 days feels pretty short to me

template.yaml
```diff
@@ -740,20 +778,69 @@ Resources:
...
+      FunctionName: !Sub "gro-publish-record-${Environment}${Developer}"
+      CodeUri: lambdas/gro-publish-record
+      Handler: uk.gov.di.data.lep.PublishRecord::handleRequest
+      Environment:
+        Variables:
+          POWERTOOLS_SERVICE_NAME: gro-publish-record
+      AutoPublishAlias: latest
+      Policies:
+        - Statement:
+            - Sid: KmsMasterKeyAccess
```
ethanmills, line 817:
```suggestion
            - Sid: KmsMainKeyAccess
```

template.yaml
```diff
@@ -685,7 +710,21 @@ Resources:
...
-            FunctionName: !Ref GroSplitFileFunction
+            FunctionName: !Ref GroConvertToJsonFunction
+        - LambdaInvokePolicy:
+            FunctionName: !Ref GroPublishRecordFunction
+        - S3FullAccessPolicy:
+            BucketName: !Ref GroRecordsBucket
+        - StepFunctionsExecutionPolicy:
+            StateMachineName: !Sub "${Environment}${Developer}-ingestion-state-machine"
+        - Statement:
+            - Sid: KmsMasterKeyAccess
```
ethanmills, line 721:
```suggestion
            - Sid: KmsMainKeyAccess
```

ethanmills (approved):

PR #1806, opened 2023-08-01
GPC-520: Get data from bucket

hannah-leong:


lambdas/gro-convert-to-json/src/main/java/uk/gov/di/data/lep/ConvertToJson.java
```diff
@@ -37,9 +37,15 @@ public ConvertToJson(AwsService awsService, Config config, ObjectMapper mapper)
     @Override
     @Tracing
     @Logging(clearState = true)
-    public GroFileLocations handleRequest(S3ObjectCreatedNotificationEvent event, Context context){
+    public GroFileLocations handleRequest(S3ObjectCreatedNotificationEvent event, Context context) {
         logger.info("Splitting file");
         logger.info("Bucket: " + event.detail.bucket.name + " has new file: " + event.detail.object.key);
```
oskarwilliams, line 42: we can probably remove/edit these now

PR #1814, opened 2023-08-02
GPC-521: Map XML file to json

hannah-leong:


lambdas/gro-convert-to-json/src/test/java/uk/gov/di/data/lep/ConvertToJsonTest.java
```diff
@@ -61,6 +67,6 @@ void convertToJsonReturnsBucketsDetails() {
     void convertToJsonUploadsToS3() {
         underTest.handleRequest(event, context);
 
-        verify(awsService).putInBucket(eq("JsonBucketName"), anyString(), eq("jsonString"));
+        verify(awsService).putInBucket(eq("JsonBucketName"), anyString(), anyString());
```
oskarwilliams, line 70: Can you verify this is the right string pls

oskarwilliams (changes requested):

PR #1815, opened 2023-08-02
GPC-NONE: Remove gds-users access to accounts

ethanmills:
More of a question than anything else. Now that we have SSO access, do we still want the gds-users roles knocking about?


oskarwilliams:
Have approved this as yep I think that is correct

PR #1817, opened 2023-08-02
GPC-509: Add xray resource policy back in

oskarwilliams:


ethanmills (approved):

ethanmills:
Out of curiosity, do multiline string like https://yaml-multiline.info/ work?

PR #1851, opened 2023-08-07
GPC-NONE: Handle DeathRegistrationGroups that have only one record

hannah-leong:


lambdas/gro-convert-to-json/src/main/java/uk/gov/di/data/lep/ConvertToJson.java
```diff
@@ -47,10 +49,12 @@ public GroFileLocations handleRequest(S3ObjectCreatedNotificationEvent event, Co
...
-            .getJSONObject("DeathRegistrationGroup")
-            .getJSONArray("DeathRegistration")
-            .toString();
+        var deathRegistrationGroup = XML.toJSONObject(xmlData).getJSONObject("DeathRegistrationGroup");
+        try {
+            return deathRegistrationGroup.getJSONArray("DeathRegistration").toString();
+        } catch (JSONException e) {
+            logger.info(e);
+            return (Arrays.toString(new Object[]{deathRegistrationGroup.get("DeathRegistration")}));
+        }
```
oskarwilliams, line 58: I think this won't be complete enough as it will mask a load of other error types, I'd suggest something like below instead. Also I would be worried about the Arrays.toString... stuff, so tests are definitely necessary for this logic :) 

```suggestion
        var deathRegistrations = deathRegistrationGroup.optJSONArray("DeathRegistration") == null 
            ? new JSONArray(List.of(deathRegistrationGroup.get("DeathRegistration")))
            : deathRegistrationGroup.getJSONArray("DeathRegistration");
        return deathRegistrations.toString();
```

oskarwilliams (changes requested):

PR #1912, opened 2023-08-14
GPC-535: Update data format for validation and enrichment lambda

oskarwilliams:


lambdas/death-enrichment/src/main/java/uk/gov/di/data/lep/GroDeathEnrichment.java
```diff
@@ -46,49 +41,25 @@ public GroDeathEventEnrichedData handleRequest(SQSEvent sqsEvent, Context contex
...
-            "2021",
-            "Bob Burt",
-            "Smith",
-            "Jane",
-            "888 Death House",
-            "8 Death lane",
-            "Deadington",
-            "Deadshire",
-            "XX1 1XX"
+            baseData.registrationId(),
```
ethanmills, line 48: I miss kotlin

lambdas/death-enrichment/src/test/java/uk/gov/di/data/lep/GroDeathEnrichmentTest.java
```diff
@@ -55,44 +57,40 @@ void constructionCallsCorrectInstantiation() {
...
+        sqsMessage.setBody("A message body");
         var sqsEvent = new SQSEvent();
         sqsEvent.setRecords(List.of(sqsMessage));
 
-        when(objectMapper.readValue(sqsMessage.getBody(), GroDeathEventBaseData.class))
-            .thenReturn(new GroDeathEventBaseData(
-                "123a1234-a12b-12a1-a123-123456789012"
-            ));
+        when(objectMapper.readValue(sqsMessage.getBody(), GroJsonRecord.class))
+            .thenReturn(new GroJsonRecordBuilder().build());
```
ethanmills, line 65: I don't _really_ care, but this feels a bit yucky. It would make more sense to me to have the enrichment (business logic) in a service class, which is called by the handler which acts like a controller and deals with details of the transport mechanism etc. You'd then be able to test it directly.

ethanmills (approved):

PR #1949, opened 2023-08-18
GPC-514: Add temporal accessor serialisation and deserialisation

oskarwilliams:


lib/src/test/java/uk/gov/di/data/lep/library/services/MapperTestObject.java
```diff
@@ -9,4 +10,5 @@ public class MapperTestObject {
     public LocalDateTime dateTime;
     public GenderAtRegistration gender;
     public GroVerificationLevel verificationLevel;
+    public IsoDate isoDate;
```
olevett, line 13: I don't think these should be fields :D

lib/src/main/java/uk/gov/di/data/lep/library/services/Mapper.java
```diff
@@ -2,18 +2,26 @@
...
+import java.time.temporal.TemporalAccessor;
+
 public class Mapper {
     private Mapper() {
         throw new IllegalStateException("Utility class");
     }
 
     public static ObjectMapper objectMapper() {
+        var temporalAccessorModule = new SimpleModule();
+        temporalAccessorModule.addSerializer(TemporalAccessor.class, new TemporalAccessorSerializer());
```
olevett, line 18: I think you should be able to add these as an annotation to the fields we want to serialize/deserialize rather than as a mapper level config
```@JsonSerialize(using =...)```
which might be more resilient

Comment (olevett) [reply lib/src/main/java/uk/gov/di/data/lep/library/services/Mapper.java:18]: Specifically, I think
```java
public record IsoDate(
    String description,
    @JsonSerialize(using = TemporalAccessorSerializer.class)
    @JsonDeserialize(using= TemporalAccessorDeserializer.class)
    TemporalAccessor value
) {
}
```

for the IsoDate works

olevett (approved):

lib/src/main/java/uk/gov/di/data/lep/library/services/Mapper.java
oskarwilliams: Instead, am making a PartialDate serializer at top level

oskarwilliams: Not doing this and doing what Olly said

lib/src/main/java/uk/gov/di/data/lep/library/services/PartialDateDeserialiser.java
```diff
@@ -0,0 +1,38 @@
...
+import java.time.temporal.TemporalAccessor;
+
+public class PartialDateDeserialiser extends JsonDeserializer<TemporalAccessor> {
+    @Override
+    public TemporalAccessor deserialize(JsonParser jsonParser, DeserializationContext deserializationContext) throws IOException {
+        var token = jsonParser.getCurrentToken();
+
+        if (token.equals(JsonToken.VALUE_STRING)) {
+            var text = jsonParser.getText().trim();
+            var countOfDashes = text.split("-", -1).length - 1;
```
olevett, line 22: Can we do the split once, then count "number of date parts" instead? (that saves us doing the same split again on ln 26)

olevett (approved):

lib/src/test/java/uk/gov/di/data/lep/library/services/MapperTest.java
```diff
@@ -74,4 +78,52 @@ void xmlMapperSerialisesXmlToObject() throws JsonProcessingException {
         assertEquals(expected.gender, result.gender);
         assertEquals(expected.verificationLevel, result.verificationLevel);
     }
+
+    @Test
+    void objectMapperMapsCompleteIsoDateCorrectly() throws JsonProcessingException {
```
olevett, line 83: These tests might make more sense on the serializers themselves, now they're not part of the Mapper config, but this is a very weak feeling.

PR #2027, opened 2023-08-25
GPC-NONE: Add fingerprint verification to GRO SFTP client

oskarwilliams:


lambdas/gro-pull-file/src/main/java/uk/gov/di/data/lep/GroPullFile.java
```diff
@@ -56,17 +56,19 @@ public GroFileLocations handleRequest(Object event, Context context) {
...
 
     private RemoteFile downloadFile() {
+        var fingerprintId = config.getGroSftpServerFingerprintSecretID();
         var hostname = config.getGroSftpServerHost();
-        var privateKeyId = config.getGroSftpServerPrivateKeySecretId();
+        var privateKeyId = config.getGroSftpServerPrivateKeySecretID();
         var sourceDir = config.getGroSftpServerSourceDir();
         var username = config.getGroSftpServerUsername();
 
+        var fingerprint = awsService.getSecret(fingerprintId);
```
ethanmills, line 65: I don't think this needs to be a secret - it should be no less public than the host details. No particular problem with doing it this way though

ethanmills (approved):

lambdas/gro-pull-file/src/main/java/uk/gov/di/data/lep/GroPullFile.java
oskarwilliams: What are your thoughts on making everything secret? So that we can configure without a redeploy

PR #2080, opened 2023-08-30
GPC-570: Setup users with permissions

hannah-leong:


template.yaml
```diff
@@ -261,6 +261,50 @@ Resources:
...
+            Resource:
+              - !Ref DwpDevDeathKmsKey
+              - !Ref DwpTestDeathKmsKey
+              - !Ref DwpStageDeathKmsKey
+
+  DwpIntegrationUserGroup:
+    Type: AWS::IAM::Group
+    Condition: IsIntegration
+    Properties:
+      GroupName: IamUserGroup
```
ethanmills, line 297: Could you give this a more meaningful name please?

template.yaml
```diff
@@ -261,6 +261,50 @@ Resources:
...
+      ManagedPolicyArns:
+        - !Ref DwpIntegrationManagedPolicy
+
+  DwpIntegrationUser:
+    Type: AWS::IAM::User
+    Condition: IsIntegration
+    Properties:
+      Groups:
+        - !Ref DwpIntegrationUserGroup
```
ethanmills, line 306: I think we should either explicitly name this user, or give it a tag which makes it clear from the console what the user is for

ethanmills (approved):

PR #2100, opened 2023-09-01
GPC-573: Replacement queue for dwp testing

hannah-leong:


template.yaml
```diff
@@ -1384,6 +1299,252 @@ Resources:
...
+        maxReceiveCount: 5
+
+  DwpOldFormatDeathMinimisationDlq:
+    Condition: IsIntegration
+    Type: AWS::SQS::Queue
+    Properties:
+      QueueName: !Sub ${Environment}${Developer}-dwp-old-format-death-minimisation-dlq
+      KmsMasterKeyId: !Ref MainKmsKey
+
+  DwpOldFormatDeathDeliveryQueue:
```
CarlyG55, line 1398: Is there a reason we've changed these names to get rid of 'dev'?

lambdas/convert-set-to-old-format/src/main/java/uk/gov/di/data/lep/ConvertSetToOldFormat.java
```diff
@@ -55,19 +55,22 @@ public String handleRequest(SQSEvent sqsEvent, Context context) {
...
-        var dateOfBirth = (LocalDate) subject.birthDate().get(0).value();
 
-        LocalDateTime deathRegistrationTime = LocalDateTime.now();
+        LocalDateTime deathRegistrationTime;
         if (events instanceof DeathRegistrationEvent registrationEvent) {
             deathRegistrationTime = registrationEvent.deathRegistrationTime().value();
         } else if (events instanceof DeathRegistrationUpdateEvent registrationUpdateEvent) {
             deathRegistrationTime = registrationUpdateEvent.recordUpdateTime().value();
+        } else {
+            throw new MappingException(new RuntimeException());
```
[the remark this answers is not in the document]
CarlyG55: If not, can we provide a message with the runtime exception?

CarlyG55 (changes requested):
I just have a couple of q's (mostly because the codebase has completely changed since I last worked on it 🙃) but looks good!

template.yaml
hannah-leong: It needs to be processed separately as otherwise it would consume the (new format) messages off the delivery queue in order to convert them to the old format - we want both new and old formats available (I think that's why)

lambdas/convert-set-to-old-format/src/main/java/uk/gov/di/data/lep/ConvertSetToOldFormat.java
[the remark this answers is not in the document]
hannah-leong: It's an extension of the RuntimeException so takes in an exception (essentially we made it just to be more specific as to what exception it is) - I'll add a message in though, good point

template.yaml
CarlyG55: sounds good!

CarlyG55 (approved):

PR #2119, opened 2023-09-04
GPC-553: Dependency scanning

hannah-leong:


hannah-leong:
Change of plans: will remove dependency-check and extend trivy instead

PR #2121, opened 2023-09-04
GPC-563: Add sensitive data tags

hannah-leong:
Add tags to data stores to indicate if they store PII/sensitive data

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -141,7 +141,7 @@ Resources:
       VersioningConfiguration:
         Status: Enabled
       Tags:
-        - Key: SensitiveData
+        - Key: PersonalIdentifyingInformation
           Value: true
 
   LambdaSecurityGroup:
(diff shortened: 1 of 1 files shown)


template.yaml
```diff
@@ -140,6 +140,9 @@ Resources:
         RestrictPublicBuckets: true
       VersioningConfiguration:
         Status: Enabled
+      Tags:
+        - Key: SensitiveData
```
CarlyG55, line 144: I'd say we probably want the tag key to be PII or PersonalIdentifyingInformation - there is actually a bit of a difference between PII and sensitive data - 

> PII is the most commonly available and least regulated type of data, and may or may not be sensitive — or may be considered sensitive only under certain circumstances, or when combined with other data about an individual.

>For example, PII like names, phone numbers, or other information that may be widely publicly available, is not usually considered sensitive (though could be in certain contexts), whereas PII like social security numbers, alien registration numbers, or driver’s license numbers would always be sensitive.

from https://bigid.com/blog/sensitive-information-guide/

template.yaml
CarlyG55, line 1: I think `DeathEnrichmentFunctionLogGroup` should have the tag - we log the registration id here

CarlyG55: Similarly `GroPublishRecordFunctionLogGroup`

CarlyG55 (changes requested):

CarlyG55 (approved):

PR #2143, opened 2023-09-06
GPC-588: Fix minimisation output

hannah-leong:


lambdas/convert-set-to-old-format/src/main/java/uk/gov/di/data/lep/ConvertSetToOldFormat.java
```diff
@@ -143,4 +144,16 @@ private static List<String> getNames(DeathRegistrationSubject subject) {
...
+
+    private LocalDate temporalAccessorToLocalDate(TemporalAccessor from) {
+        if (from.isSupported(ChronoField.YEAR) && from.isSupported(ChronoField.MONTH_OF_YEAR) && from.isSupported(ChronoField.DAY_OF_WEEK)) {
+            return LocalDate.from(from);
+        } else if (from.isSupported(ChronoField.YEAR) && from.isSupported(ChronoField.MONTH_OF_YEAR)) {
+            return LocalDate.of(from.get(ChronoField.YEAR), from.get(ChronoField.MONTH_OF_YEAR), 1);
+        } else if (from.isSupported(ChronoField.YEAR)) {
+            return LocalDate.of(from.get(ChronoField.YEAR), 1, 1);
+        } else {
+            return LocalDate.of(2015, 1, 1);
```
olevett, line 156: This feels like we should throw rather than return a completely made up date!

hannah-leong: Because this is only test data for the old format, figured it would be okay to just sub in a made up date

CarlyG55: Is it just test data? It's not in the test folder?

lambdas/death-enrichment/src/test/java/uk/gov/di/data/lep/DeathEnrichmentTest.java
```diff
@@ -35,6 +37,10 @@ class DeathEnrichmentTest {
     private static final ObjectMapper objectMapper = mock(ObjectMapper.class);
     private static final DeathEnrichment underTest = new DeathEnrichment(awsService, config, objectMapper);
 
+    @BeforeAll
+    static void setup(){
+        when(config.getTargetTopic()).thenReturn("Target Topic");
+    }
     @BeforeEach
```
olevett, line 44:
```suggestion
    }
    
    @BeforeEach
```

lambdas/convert-set-to-old-format/src/main/java/uk/gov/di/data/lep/ConvertSetToOldFormat.java
hannah-leong: Sorry, its data that is generated - I think the idea was that it was just so that dwp could test that it worked/they could get data from the queues; i.e., not being used with real data therefore only important to get the structure of data rather than the actual content.

olevett: I worry about that as an assumption, this could slip through. I also think that in the scenario where there is no date data supplied, we should be erroring or passing through null explicitly.

CarlyG55: I just chatted to Ethan and I'm convinced it's alright because it's only deployed in integration, there's a condition to stop it being deployed in prod

CarlyG55: @olevett ethan said if you'd like to have an argument with him he's happy to talk about it

olevett: Okay, fine...I'm not sure I know why just returning null here is a problem, but sure

PR #2162, opened 2023-09-08
GPC-480: Add data retention policy

CarlyG55:


doc/data-retention-policy.md
```diff
@@ -0,0 +1,27 @@
+# Data Retention Policy
+
+We want to prevent the loss of data when there are processing issues, whilst minimising the amount of time we hold any
+data for.
+
+## Maximum Retention Periods
+
+All of our source queues have a maximum retention period of 4 days. This is to allow for any processing issues to be
```
galund, line 8: For what it's worth I'd be inclined to make this 7 days to account for the sometimes 4-5 days of developer non-availability over some bank holiday periods eg Christmas.

doc/data-retention-policy.md
```diff
@@ -0,0 +1,27 @@
...
+identified in the event of problems over bank holidays, or periods of low team availability.
+
+Our dead letter queues have a retention period of 14 days, to maximise the time allowed to fix any issues causing
+message delivery to fail.
```
galund, line 12:
```suggestion
message delivery (or processing) to fail.
```

galund (commented):
Our data retention policy also needs to cover audit logging and anything else external. Although that's still being discussed, so this can land without, but a placeholder for that (or  a statement of intent about what we're aiming for there) might be a good idea.

There are some directorate-level artefacts that we should probably be working with Dan Budden to get update, although he might decide that it's best to keep Life Events stuff separate for now - see https://govukverify.atlassian.net/wiki/spaces/Architecture/pages/3652780947/Data+retention

doc/data-retention-policy.md
CarlyG55: Sure - I had a little bit of discussion about this with Ethan and Olly, and we were discussing 4 days vs 7 days, but went with 4 days mostly because that's the number we'd already given to DWP, and that 4 days of developer non-availability is rare. But I think it probably makes sense to go with 7, and we can just tell DWP we've changed it

CarlyG55:
@galund with regards to adding to that directorate-level data retention sheet - I wouldn't be surprised if they want to keep it separate, as it feels like our data is quite different to the ones in that sheet. I only have access to the general access sheet though, not the limited access, so I'm not sure if it looks much different. Should I send a message over to Dan, or did you want to do that?

galund:
> Should I send a message over to Dan, or did you want to do that?

I've emailed 👍

olevett:
This looks good to me, but if this forms part of external contract with other departments, I wonder if we want it in TechDocs rather than just the internal docs?

olevett (approved):

CarlyG55:
(realised I moved this into tech docs in the wrong way - fixing now)

techdocs/source/data-retention-policy.html.md.erb
```diff
@@ -0,0 +1,36 @@
+---
+title: Onboarding instructions for Acquirers
```
olevett, line 2:
```suggestion
title: Data Retention Policy
```

PR #2217, opened 2023-09-13
GPC-491: Xml files for Gro data edge case testing

hannah-leong:


hannah-leong:
Files that succeed:
- characterCount.xml; note, max character count is not enforced in our lambdas
- fieldOccurances.xml; note, max occurances are not enforced in our lambdas
- nonAsciiCharacters.xml
- updatedRecords.xml

Files that do not currently succeed:
- noDeathRegistrations (fails in state machine); add handling to convertToJson to check number of records and if none, do not try to map
- onlyRequiredFields (all records fail at enrichment); add handling for optional fields (e.g. given name, postcode)

olevett (approved):

PR #2228, opened 2023-09-14
GPC-611: Handle optional fields

hannah-leong:


CarlyG55 (approved):

lib/src/main/java/uk/gov/di/data/lep/library/dto/deathnotification/DeathNotificationSetMapper.java
```diff
@@ -175,12 +175,39 @@ private static String getAliasNameTypeOrNull(List<String> deceasedAliasNameTypes
...
             );
-        return new Name(
-            description,
-            Stream.concat(givenNameParts, Stream.of(new NamePart(NamePartType.FAMILY_NAME, nameStructure.personFamilyName()))).toList()
-        );
+
+        if (description == null) {
+            return new Name(
+                null,
+                givenNameParts == null
```
CarlyG55, line 185: This feels a bit repeated to me (multiple checks if given names are null) but I'm not entirely convinced there's a good way around that

CarlyG55: I think it's probably just me being sad there's no nullish coalescing operator 🥲

CarlyG55 (approved):

PR #2252, opened 2023-09-18
GPC-584: Stop failing silently if no recordLocked/Update time 

hannah-leong:


CarlyG55 (commented):
Hmm - I can't help but feel like this check should be done in validation. If we throw an error in validation, the user knows they've given incorrect input, but if we throw it in enrichment, it's too late...

hannah-leong:
> Hmm - I can't help but feel like this check should be done in validation. If we throw an error in validation, the user knows they've given incorrect input, but if we throw it in enrichment, it's too late...

I thought this too and had a chat with Olly about this last week regarding the other edge cases - I think we decided to have it in enrichment because validation is simply to check that the format that the data is coming in is valid (i.e., for if we get direct events from HMPO rather than process the files, we need to check what they send is in the correct layout)

CarlyG55:
> I thought this too and had a chat with Olly about this last week regarding the other edge cases - I think we decided to have it in enrichment because validation is simply to check that the format that the data is coming in is valid (i.e., for if we get direct events from HMPO rather than process the files, we need to check what they send is in the correct layout)

Which cases were they? I'd argue that this one actually is that the user is giving data in the wrong format (although we also need to update our docs to confirm that)! If data doesn't have either datetime, it is in the incorrect format but it's hard to check that in the schema check, and has to be done a bit more manually like this

lambdas/death-validation/src/main/java/uk/gov/di/data/lep/DeathValidation.java
```diff
@@ -43,7 +43,15 @@ private GroJsonRecord validateRequest(APIGatewayProxyRequestEvent event) {
...
         try {
-            return objectMapper.readValue(event.getBody(), GroJsonRecord.class);
+            var groJsonRecord = objectMapper.readValue(event.getBody(), GroJsonRecord.class);
+            if (groJsonRecord.recordLockedDateTime() == null && groJsonRecord.recordUpdateDateTime() == null) {
+                logger.error("Record has neither recordLocked and recordUpdate dateTimes");
+                throw new MappingException("Record has neither recordLocked and recordUpdate dateTimes");
+            } else if (groJsonRecord.recordLockedDateTime() != null && groJsonRecord.recordUpdateDateTime() != null) {
+                logger.error("Record has both recordLocked and recordUpdate dateTimes");
+                throw new MappingException("Record has both recordLocked and recordUpdate dateTimes");
+            }
```
CarlyG55, line 53: do we definitely only want one of these?

template.yaml
CarlyG55, line 1: weird these changes have come up - I'm going to assume you haven't done anything to this

hannah-leong: Yeah, it was after I rebased for some reason :/

lambdas/death-validation/src/main/java/uk/gov/di/data/lep/DeathValidation.java
hannah-leong: According to the schema, it is an either or case. We are, however, waiting for confirmation.

ethanmills (approved):

PR #2255, opened 2023-09-18
Copy data model RFC into tech docs

galund:
I'm not sure this is a good idea...


ethanmills:
I've got an alternative here: https://github.com/alphagov/di-data-life-events-platform/compare/build-rfc-docs-alt?expand=1 

I tend to avoid submodules where I can, but not a hill I'll die on

ethanmills:
I've also opened https://github.com/alphagov/di-data-life-events-platform/pull/2257 to copy it in alone while we nitpick over which horrible method is less horrible 😁

PR #2272, opened 2023-09-19
GPC-574: send audit data to queues

CarlyG55:


lambdas/death-enrichment/src/main/java/uk/gov/di/data/lep/DeathEnrichment.java
```diff
@@ -42,9 +48,15 @@ public String handleRequest(SQSEvent sqsEvent, Context context) {
     }
 
     @Tracing
-    private DeathNotificationSet enrichData(GroJsonRecord baseData) {
-        logger.info("Enriching and mapping data (sourceId: {})", baseData.registrationID());
+    private DeathNotificationSet enrichData(GroJsonRecordWithCorrelationId baseData) {
+        logger.info("Enriching and mapping data (sourceId: {})", baseData.groJsonRecord().registrationID());
```
olevett, line 52: I think given the comments about sensitivity it might be worth using the correlation id here instead of the registration ID

lambdas/death-minimisation/src/main/java/uk/gov/di/data/lep/DeathMinimisation.java
```diff
@@ -43,6 +45,10 @@ public String handleRequest(SQSEvent sqsEvent, Context context) {
             var minimisedData = objectMapper
                 .writer(filterProvider)
                 .writeValueAsString(enrichedData);
+
+            var auditData = generateAuditData(minimisedData, enrichedData.txn());
```
olevett, line 49: I'd be half tempted to shift the generation and the auditing into their own method, as `auditMinimsation` or something similar - we don't need the auditData return value for anything else

olevett: This would apply to the other cases too

lambdas/gro-convert-to-json/src/main/java/uk/gov/di/data/lep/GroConvertToJson.java
```diff
@@ -107,11 +113,13 @@ private String convertXmlDataToJson(String xmlData, String authorisationToken) {
...
             validateRecordCount(records, deathRegistrationGroup.recordCount());
 
-            var recordsWithAuth = records.stream()
-                .map(r -> new GroJsonRecordWithAuth(r, authorisationToken))
+            var recordsWithHeaders = records.stream()
+                .map(r -> new GroJsonRecordWithHeaders(r, authorisationToken, UUID.randomUUID().toString()))
                 .toList();
 
-            return objectMapper.writeValueAsString(recordsWithAuth);
+            recordsWithHeaders.forEach(record -> addAuditDataToQueue(generateAuditData(record.correlationId(), xmlData.hashCode())));
```
olevett: I have a couple of thoughts on this
- I find .forEach a little confusing as a method (personal opinion :D) because it looks like it's a functional side effect-less method call, but it's actually only about side effects
- I think if we can shift the mapping from a list of events to individual audit calls down into a child method, we might be able to take advantage of the SQS Batch API which could save us a lot of overhead

Happy to talk about this though

lambdas/death-enrichment/src/main/java/uk/gov/di/data/lep/DeathEnrichment.java
CarlyG55: Yeah, I was wondering this - so all logs we want correlation id?

lambdas/death-minimisation/src/main/java/uk/gov/di/data/lep/DeathMinimisation.java
CarlyG55: Yeah I think it was because it was a lot more doable to commonise the auditing rather than the generating I went for this, but can have another look at it

lambdas/gro-convert-to-json/src/main/java/uk/gov/di/data/lep/GroConvertToJson.java
CarlyG55: Oh nice! Yeah sure, happy to look at this

lambdas/death-minimisation/src/main/java/uk/gov/di/data/lep/DeathMinimisation.java
olevett: Probably not worth spending too much time on this, I think it's very marginal!

lambdas/death-enrichment/src/main/java/uk/gov/di/data/lep/DeathEnrichment.java
olevett: I think where we've got a correlation ID we should use it, and at at least one point we need to log the correlation id with a source ID so we can link the two in the rest of the audit trail

lambdas/gro-convert-to-json/src/main/java/uk/gov/di/data/lep/GroConvertToJson.java
CarlyG55: Discussed with Ethan - SQS Batch API has a limit of 10 messages, and there doesn't seem to be a particularly nice way to split up the messages in Java like this without a new package. SQS doesn't have a rate limit, so going to see what this looks like, and if it doesn't look doable, I'll add a dependency to split up the list more nicely.

CarlyG55: I've pulled into a child method, I can't decide if it's much nicer 😆 but I don't really mind either way

lambdas/death-enrichment/src/main/java/uk/gov/di/data/lep/DeathEnrichment.java
CarlyG55: Sure - do we think think that the DeathEnrichment logs now contain PII, if they just have the correlationID? As we've marked them as PII in the template currently

ethanmills (approved):

PR #2341, opened 2023-09-29
GPC-602: Internal queue alerts

hannah-leong:


template.yaml
```diff
@@ -458,8 +458,241 @@ Resources:
...
+      Tags:
+        - Key: CheckovRulesToSkip
+          Value: CKV_AWS_26
+
+  AlertTopicSubscription:
+    Type: AWS::SNS::Subscription
+    Properties:
+      Endpoint: di-life-events-platform@digital.cabinet-office.gov.uk
+      Protocol: email
+      TopicArn: !Ref AlertTopic
```
CarlyG55, line 479: Think we might want `!Ref AlertTopic.TopicArn`

template.yaml
```diff
@@ -458,8 +458,241 @@ Resources:
...
+      Statistic: Sum
+      Period: 60
+      EvaluationPeriods: 1
+      Threshold: 0
+      ComparisonOperator: GreaterThanThreshold
+      TreatMissingData: notBreaching
+      AlarmActions:
+        - !Ref AlertTopic
+
+  OneHourOnAuditQueueAlarm:
```
CarlyG55, line 589: I'd be reluctant to put an alarm in for the audit queue, since we don't know what we're doing with the data on the audit queue atm (unless something has changed whilst I've been away!)

CarlyG55: Still worth keeping the audit dlq alarm though

template.yaml
```diff
@@ -458,8 +458,241 @@ Resources:
...
+      Statistic: Sum
+      Period: 60
+      EvaluationPeriods: 5
+      Threshold: 3600
+      ComparisonOperator: GreaterThanThreshold
+      TreatMissingData: notBreaching
+      AlarmActions:
+        - !Ref AlertTopic
+
+  OneHourOnDeathEnrichmentQueueAlarm:
```
CarlyG55, line 607: Sorry these alarms have now changed to one minute after ethan compared it to our slo's

template.yaml
```diff
@@ -458,8 +458,241 @@ Resources:
...
+    Properties:
+      AlarmName: one-hour-on-death-minimisation-queue-alarm
+      Namespace: AWS/SQS
+      MetricName: ApproximateAgeOfOldestMessage
+      Dimensions:
+        - Name: QueueName
+          Value: !GetAtt DwpDeathMinimisationQueue.QueueName
+      Statistic: Sum
+      Period: 60
+      EvaluationPeriods: 5
```
CarlyG55, line 636: Is there a reason you went for 5 here? I think I'd probably want 1 or 2, as this just extends the time before we actually get the alarm

template.yaml
CarlyG55: note the delivery ones are still the same though

template.yaml
hannah-leong: Pretty sure in this case !Ref gets the ARN - I tested this with my personal email and it worked.

PR #2344, opened 2023-10-02
GPC-621: use sha hashing

CarlyG55:


lambdas/death-enrichment/src/main/java/uk/gov/di/data/lep/DeathEnrichment.java
```diff
@@ -55,7 +56,13 @@ private DeathNotificationSet enrichData(GroJsonRecordWithCorrelationID baseData)
 
     @Tracing
     private void audit(DeathNotificationSet enrichedData) {
-        var auditDataExtensions = new DeathEnrichmentAuditExtensions(enrichedData.hashCode(), enrichedData.txn());
-        addAuditDataToQueue(new DeathEnrichmentAudit(auditDataExtensions));
+        try {
+            var enrichedDataHash = Hasher.hash(objectMapper.writeValueAsString(enrichedData));
```
CarlyG55, line 60: I'm not sure I'm a massive fan of this approach (as we're writing the value as string twice), but hashing an object seemed more fiddly and maybe the less correct way of doing it?

lambdas/death-enrichment/src/test/java/uk/gov/di/data/lep/DeathEnrichmentTest.java
```diff
@@ -49,6 +51,7 @@ static void setup(){
     @BeforeEach
     void refreshSetup() {
         reset(objectMapper);
+        clearInvocations(awsService);
```
CarlyG55, line 54: we seem to use both reset and clearInvocations in the codebase - probably want a consistent approach, and I think that's clearInvocations for awsService, but wanted to check before updating them all

olevett: I'm not sure which is better here, but I think lets split fixing this/making it consistent somewhere else/in a different ticket.

They do achieve different things, so it's not completely unreasonable for us to have both.

olevett (approved):

PR #2350, opened 2023-10-03
GPC-612: Remove PII from step function

hannah-leong:


lambdas/gro-convert-to-json/src/main/java/uk/gov/di/data/lep/GroConvertToJson.java
```diff
@@ -146,4 +150,18 @@ private void generateAndAddListOfAuditDataToQueue(List<GroJsonRecordWithHeaders>
...
         }
     }
+
+    @Tracing
+    private String uploadIndividualRegistrations(String bucket, List<GroJsonRecordWithHeaders> deathRegistrations) throws JsonProcessingException {
+        var jsonKeys = new ArrayList<RecordLocation>();
+        for (var registration : deathRegistrations) {
+            var registrationJsonKey = registration.correlationID() + ".json";
+            jsonKeys.add(new RecordLocation(bucket, registrationJsonKey));
+            awsService.putInBucket(bucket, registrationJsonKey, objectMapper.writeValueAsString(registration));
```
olevett, line 160: What happens if one of these fails?

hannah-leong: In what sense?
They get mapped before and so each registration should be a valid GroJsonRecordWithHeaders - if it isn't then it throws an exception during the converting from xml to json.
If the mapping fails then it throws a JsonProcessing error which gets caught and thrown as a MappingException in the handleRequest function

lambdas/gro-convert-to-json/src/main/java/uk/gov/di/data/lep/GroConvertToJson.java
```diff
@@ -146,4 +150,18 @@ private void generateAndAddListOfAuditDataToQueue(List<GroJsonRecordWithHeaders>
...
+        }
+        var jsonKey = UUID.randomUUID() + ".json";
```
ethanmills, line 162: I think I'd prefer these to be deterministic. If the function times out, or there's a hiccup in S3 or something else that causes this function to error, we'd be left with a load of orphan objects. An S3 put is an upsert, so if the filenames were consistent then a retry would succeed (assuming the original issue was resolved) and the files would be cleaned up properly

olevett (approved):

PR #2354, opened 2023-10-04
GPC-624: Add new parameter for GitHub migration

olevett:


olevett:
I've manually applied this change locally, so this isn't strictly required, but made sense to track. There'll be a follow up action to remove the old GitHub repo when this is done too.

ethanmills (approved):

PR #2360, opened 2023-10-05
GPC-631: Sign lint fixed commits

olevett:


.github/workflows/update-data-model.yml
```diff
@@ -20,10 +20,17 @@ jobs:
...
 
+      - name: Import GPG key
+        uses: crazy-max/ghaction-import-gpg@v6
+        with:
+          gpg_private_key: ${{ secrets.COMMIT_BOT_GPG_KEY }}
+          passphrase: ${{ secrets.COMMIT_BOT_GPG_PASSPHRASE }}
+          git_user_signingkey: true
+          git_commit_gpgsign: true
+
       - uses: peter-evans/create-pull-request@v5
```
olevett, line 31: I'm not totally sure if this solves the problem/we need it quite like this, but thought it would be worth starting with and seeing if it fails

ethanmills (approved):

.github/workflows/update-data-model.yml
```diff
@@ -20,10 +20,19 @@ jobs:
           ARCH_TOKEN: ${{ secrets.ARCHITECTURE_READONLY_PAT }}
         run: scripts/update-rfc-techdocs.sh
 
+      - name: Import GPG key
+        uses: crazy-max/ghaction-import-gpg@v6
```
olevett, line 24:
```suggestion
        uses: crazy-max/ghaction-import-gpg@82a020f1f7f605c65dd2449b392a52c3fcfef7ef
```

PR #2361, opened 2023-10-05
GPC-NONE: Remove deprecated deploy parameter

olevett:


ethanmills (approved):
Out of curiosity, are you planning on following this up and removing the parameter entirely once removed?

olevett:
I _think_ so, but I wanted to do it in a couple of stages, because I couldn't work out quite what was going to happen :)

PR #2366, opened 2023-10-06
GPC-NONE: Update link to glossary

olevett:


olevett:
This will currently redirect back to the docs.life-events.dev.account.gov.uk domain, I wonder if we should tweak that at some point in future

CarlyG55 (approved):

PR #2387, opened 2023-10-10
GPC-641: Add state machine execution history logging

hannah-leong:


template.yaml
```diff
@@ -2390,6 +2390,12 @@ Resources:
                 bucket:
                   name:
                     - !Ref GroIngestionBucket
+      Logging:
+        Destinations:
+          - CloudWatchLogsLogGroup:
+              LogGroupArn: !GetAtt GroIngestionStateMachineLogGroup.Arn
+        IncludeExecutionData: true
```
CarlyG55, line 2397: I think we want this to be false? To avoid logging all of the PII going through (again 🙃 )

hannah-leong: It seems to be only data about the state machine execution states (i.e., task state entered, task state exited). I've checked the logs and I can't find any PII so I think it should be okay?
image

CarlyG55: Lol I forgot that we've implemented the removing PII from step functions already 🙃 if we hadn't it would be in the input/output, but since we've done that change this is fine

CarlyG55 (approved):

PR #2388, opened 2023-10-10
GPC-644: Add MFA delete for logs bucket

hannah-leong:


ethanmills:
I'm not sure this will make security hub happy? This isn't "enabling MFA delete" exactly, which isn't possible using cloudformation: https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/423

hannah-leong:
> I'm not sure this will make security hub happy? This isn't "enabling MFA delete" exactly, which isn't possible using cloudformation: [aws-cloudformation/cloudformation-coverage-roadmap#423](https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/423)
> 
> See https://docs.aws.amazon.com/AmazonS3/latest/userguide/MultiFactorAuthenticationDelete.html for what I would expect this change to involve

This was an ITHC issue, rather than security hub, so as MFADelete can't be done through cloudformation I took this route (the ITHC just mentions that MFA should be used, not specifying MFADelete).
That being said, I realise I haven't applied it to all the areas that are affected so will implement that now

CarlyG55:
I've chatted to Ethan about it, and we think it's probably not worth doing this - as to do anything we realistically need to be logged in with MFA anyway, vs asking for MFA directly when we delete a file. This just adds an extra permission that might be confusing for anyone coming back to this in the future. If we were to do this in the future, we'd have a script to run that would pull all of the bucket names, and from those run a cli command to set MFA delete, but given the situation it's not really worth doing that. Sorry!

PR #2389, opened 2023-10-10
GPC-640 VPC flow logging

CarlyG55:


template.yaml
```diff
@@ -156,7 +156,7 @@ Resources:
         ServerSideEncryptionConfiguration:
           - ServerSideEncryptionByDefault:
               SSEAlgorithm: aws:kms
-              KMSMasterKeyID: !Ref MainKmsKey
+              KMSMasterKeyID: !GetAtt MainKmsKey.Arn
```
CarlyG55, line 159: Have checked and this just update the bucket, doesn't destroy & create

PR #2393, opened 2023-10-11
GPC-646: Ignore potential vulnerability

olevett:
Not looked at this in much detail yet, but there doesn't appear to be a fix available yet through the AWS dependencies that pull this in


ethanmills:
I think there's no vuln here given we're not serving traffic? Seems fine to ignore

ethanmills (approved):

PR #2395, opened 2023-10-11
GPC-642: Add new least-privilege security group (VpcSecurityGroup)

hannah-leong:
Cannot delete default security groups but can delete the rules that allow all traffic - this needs to be done through the console (cannot be done through cloudformation).

Have added new least-privilege security group into the template and have deleted the default security ingress/egress rules (https://docs.aws.amazon.com/securityhub/latest/userguide/ec2-controls.html#ec2-2)


ethanmills:
Do we want one of these for ingress too?

